PR #116, opened 2014-06-03
Add ssl options

jonathanrico:
Adding support for setting ssl options in case we need to specify client certificates or disable peer verification


timrogers:
@jonathanrico Thanks!

PR #126, opened 2014-08-04
Allow hashie 2.*

Drakula2k:


greysteil:
As with https://github.com/ejholmes/restforce/pull/147 this can be closed now.

PR #127, opened 2014-08-13
Allow hashie 2.1.x, test suite passes

p-han:
Requiring hashie < 2.1 creates a conflict with grape 0.8.0 which requires hashie >= 2.1.


greysteil:
As with https://github.com/ejholmes/restforce/pull/147 this can be closed now. :dancer:

PR #130, opened 2014-09-05
Make logger a config option

DanOlson:
I'd really like to be able to configure Restforce to use `Rails.logger` rather than logging to STDOUT. Then in my various environments I can have Restforce logs alongside my Rails logs. It would really help for debugging.


timrogers:
Using the similar #145 - but thanks for the PR!

PR #132, opened 2014-09-10
Dupe attributes in `update!`

malclocke:
Previously the `id` was removed from the hash in the caller.


timrogers:
Fixed by #185.

PR #140, opened 2014-09-29
Fix Typo

kris-luminar:
describe_layouts, note the plural


timrogers:
@kris-luminar Thanks for this - not merging, but will implement this now.

PR #147, opened 2014-11-03
Allow Hashie 3 or lower

tonyla:
I use a gem that requires Hashie > 3.0 and it is causing a gem conflict with RestForce.

I upgraded the dependency to allow for 3.0 and ran the specs and they run clean. I also when through the list 
https://github.com/intridea/hashie/blob/master/UPGRADING.md and RestForce does not seem to be affected by any of the Hashie 3 changes.


greysteil:
@timrogers - this can be closed now. The gemspec was relaxed to `['>= 1.2.0', '< 4.0']` :dancer:

PR #149, opened 2014-11-06
Allows logger to be configured.

mbarackman-leaf:
Creates getter and setter for @logger on Restforce.configuration.


timrogers:
Using one of the other similar PRs for this. Thanks @mbarackman-leaf!

PR #158, opened 2014-12-15
Add query options

ryansch:
Docs: http://www.salesforce.com/us/developer/docs/api_rest/Content/headers_queryoptions.htm

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -114,9 +114,10 @@ module Restforce
         query('select id from Organization').first['Id']
       end
 
-      # Public: Executs a SOQL query and returns the result.
+      # Public: Executes a SOQL query and returns the result.
       #
       # soql - A SOQL expression.
+      # options - Hash of options to control the query
       #
       # Examples
       #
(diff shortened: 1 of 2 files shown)


ryansch:
I've added some additional features to allow for complete manual pagination.

ryansch:
As per our earlier ruby version discussion, ruby 1.9.3 is now past end of life.

ryansch:
We've been using this in production for over a month now.

theSteveMitchell:
@ryansch Can you help explain a little more what we're trying for here?

It looks like you're trying to take the next_page URL out of Restforce::Collection and exposing the next_page_url as something the client can control.  What's the benefit of this?

Restforce::Collection already includes enumerable so you can use it as is in the following ways:

```ruby 
collection = client.query('select Name from Account', batch_size: 1)
collection.each do | account |
    puts account.Name
end
```
or:
```ruby 
collection = client.query('select Name from Account', batch_size: 1)
collection.each_slice(1_000) do | accounts |
    puts accounts.map(&:Name)
end
```

Collection does its own fetching, so I don't get why we would want to remove that abstraction.

ryansch:
We use sidekiq to split up all of our tasks.  In this case I'm using a sidekiq job per page to fetch data from salesforce.  I don't want large/long running jobs in our async environment so I need access to the individual pages (and url) so I can queue them up.

theSteveMitchell:
Interesting.  IIRC the url includes a query locator and an offset, and you can only get the fetch url for the next page, one at a time.

Does that mean your Sidekiq jobs are created in serial?  One job pulls a page of results and then passes the next fetch url to a new job?  I think we could accomplish the same by just passing the Restforce::collection to the new job.  I know Sidekiq doesn't recommend serializing complex objects as parameters, but that's kind of what enumerable is for.

Somewhat separate issue, since salesforce has to keep the query locator open in order to fetch results, it feels more like you should do a long running job anyway to avoid contention there.

Would like @ejholmes to weigh in on this, it doesn't seem to me like a common use case we want to support in master, as it makes the API a lot more complicated and easy to misuse.

ryansch:
We queue the next page (if there is one) before we handle the records in the current page.  This causes all of the pages to be fetched before the query locator can time out.  I _can't_ use long running jobs in sidekiq as the entire system is designed around very quickly running jobs.

I don't see how this is a large change to the API.  I'm exposing a next_page_url that doesn't need to be used if you want the auto-paginating collection behavior.  I'm also exposing an optional options hash for query and adding a direct way to query the next page.  None of this changes the way restforce is used for the general case but it exposes enough functionality to allow for individual page handling.

PR #162, opened 2014-12-19
Add authentication_wrapper

ryansch:
Adds support for an authentication wrapper that runs around the
authentication action.  The callable should call/yield to the passed
block to run the authentication.

Note: This PR requires ruby 2.0 or greater.  See discussion on #158 about ruby versions.


ryansch:
We've been using this in production for over a month now.

timrogers:
Won't be using this as we're going Ruby 1.9.3 or later for now.

Feel free to makea new PR with a 1.9.3-compatible version - although it'd be good to have more of an explanation of why this is useful as it looks like it might be quite specific to your use case.

PR #163, opened 2014-12-19
Instruct travis to test newer rubies

ryansch:
After starting a discussion in #158 about ruby versions I figured I should have a proper PR to continue that discussion.  The only thing I see here is that we lose jruby-19mode in favor of jruby-head.  I'm not a jruby user myself so I would be interested in hearing some opinions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,9 @@
 language: ruby
 rvm:
-  - 1.8.7
-  - 1.9.2
-  - 1.9.3
   - 2.0.0
-  - 2.1.2
-  - jruby-18mode
-  - jruby-19mode
-  - ree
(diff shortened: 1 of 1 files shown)


timrogers:
FYI, I'm joining @ejholmes on maintaining the gem now. 

Ruby 1.8.7 has been end-of-lifed now and being stuck on it prevents us from taking advantage of nice things like hash syntax. Few use 1.8.7 now. I'll plan to release a new major version soon with >1.9.3 support only.

theSteveMitchell:
:+1:

timrogers:
Merged in favour of https://github.com/ejholmes/restforce/pull/189.

PR #168, opened 2015-01-15
Allow blocks to pass through

jxa:
`Restforce.new` was dropping block arguments when calling through to the
underlying `Data::Client`.


ryansch:
:+1: Just hit this.

ryansch:
We've been using this in production for almost a month now.

jxa:
We've been using in production since Jan 15. @ejholmes how do you feel about merging?

timrogers:
:+1: Where was this causing problems before? What was the use case?

PR #172, opened 2015-02-17
Paul Nunnerley - Support for data replication end points

paulnunnerley:
We've needed to extend to provide support for the data replication end
points. Doing so means upgrading the api version to 29+ and implementing
two end points getUpdatedBetween and getDeletedBetween which take two
Time object representing the beginning and end of a time window you wish
salesforce to provide you notification of updated. 

I've pulled in active support for :test group to provide the 10.minutes style syntax.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,9 +3,6 @@ gemspec
 
 gem 'rake'
 gem 'jruby-openssl', :platforms => :jruby
-group :test do
-  gem 'activesupport'
-end
 
 group :development do
   gem 'guard-rspec'
--- a/spec/unit/concerns/api_spec.rb
+++ b/spec/unit/concerns/api_spec.rb
@@ -1,5 +1,4 @@
 require 'spec_helper'
-require 'active_support/all'
 
 describe Restforce::Concerns::API do
   let(:response) { double('Faraday::Response', :body => double('Body')) }
@@ -234,8 +233,8 @@ describe Restforce::Concerns::API do
 
   describe '.get_updated_between' do
       let(:sobject){ 'Account'}
-      let(:startTime){ Time.now }
-      let(:endTime){ Time.now - 10.minutes }
+      let(:endTime){ Time.now  }
+      let(:startTime){ Time.now - 100}
       subject(:result){ client.get_updated_between(sobject, startTime, endTime)}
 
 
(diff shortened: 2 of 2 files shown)


timrogers:
Will take a look at this after the stuff to deal with versions in https://github.com/ejholmes/restforce/pull/178 is decided.

timrogers:
@paulnunnerley Happy to merge this, but would prefer to stick with the existing API version and use the `version_guard` approach from #178. Can you write a README entry too, please?

timrogers:
@paulnunnerley Going to hold off on implementing this one for now - it'd be great to have (a) a README entry and (b) an implementation which is a bit more agnostic as to what the `startDate` and `endDate` attributes should be - calling `utc.iso8601` isn't particularly [duck-typing](http://rubylearning.com/satishtalim/duck_typing.html) friendly.

BradRuderman:
+1, I need this functionality too. Thanks!

kirantpatil:
Hi, Any updates on this MR ?

jonydev:
Hi, could this pull request be merged soon? We are facing the same problem.

timrogers:
@jonydev Happy to see this merged with the changes I suggested in my comments above. Do you want to make those tweaks? I'd prefer not to do it as I'm not very familiar with these Salesforce API features, so I won't do the best job.

jonydev:
hi, @timrogers , do you mean I should add some version_guard in README.md?

timrogers:
I'll comment back in the next day or so to make a bit clearer what I'm
imagining as the next steps.

timrogers:
@jonydev It'd be good to make an updated version of this PR which:

* Uses `version_guard` to only make it available in supported API versions
* Includes a README entry
* Passes the start date and end_date in a nicer way - calling `.utc.iso8601` on it is a bit too specific. It might be better just to take a formatted string?

jonydev:
@timrogers I will add all guide you mentioned, and create a new PR for the modification. Thanks for your response.

jonydev:
@timrogers I have seen the latest version of RestForce, I found there has been an API called "get_updated", "get_deleted" did the same thing with the code we post. So I think this PR could be closed.

timrogers:
Awesome! Let's do that then. Thanks.

PR #177, opened 2015-05-01
Add limits resource to api

zenchild:
Adds the **limits** resource to api.rb.
This also adds a `version_guard` method that will raise an exception if you aren't connected at a version that supports this feature. It was written in a way that can be re-used for other resources.

https://www.salesforce.com/us/developer/docs/api_rest/Content/dome_limits.htm


timrogers:
Closing this as the same stuff appears in the other PR.

Issue #182, opened 2015-05-11
Custom Fields not accessible


demonmind:
When i use a custom fields i always get:
`Faraday::ResourceNotFound: NOT_FOUND: Provided external ID field does not exist or is not accessible: FieldName`

tried this: 
`sf_account = house_client.select("Account", 'external_id', ['Id'], 'FieldName__c')`
`sf_account = house_client.find("Account", 'external_id', 'FieldName__c')`

theSteveMitchell:
Are you trying to query by a custom field, or trying to query by Id, retrieving a custom field?

if it's the latter, you might want to try
```sf_account = house_client.select("Account", 'external_id', ['FieldName__c'])```

The third parameter is an array of fields to retrieve.  The fourth is an optional External_id field (so it lets you query BY this field, instead of the default Id column)

So to use a more concrete example:

```client.select('Account', '001D000000INjVe', ["Id"], 'Some_External_Id_Field__c'```
Will essentially do "select Id from Accounts where Some_External_Id_Field__c = '001D000000INjVe' "

Whereas
```sf_account = house_client.select("Account", 'external_id', ['Id'], 'FieldName__c')```
Will essentially be equivalent to "select Id from Account where FieldName__c = 'external_id' "

To get the best performance out of this, there is an additional caveat that 'FieldName__c' must ben an "external id field"  What that means is that in Salesforce must have this attribute attribute to make sure it is always unique and indexed properly.  The error message you're getting means that FieldName__c is not an external id column, so you can't query by it.  

If you're trying to query BY FieldName__C, you could just use

```sf_account = house_client.query(select * from Account where FieldName__C == 'asdlfkjadsflk')```

Lots of words.  Hope this helps.  Let me know if you have a suggestion to make this clearer in the docs.

demonmind:
Thank you for clarifying this. I ended up using the `.query` method and writing the SOQL query but this will be helpful in the future.

LucasCioffi:
@demonmind to create a `external ID field` is there a specific data type one must?  In the screenshot below, `External Lookup Relationship` comes closest but that's probably not correct, because on the next Salesforce screen I get an empty dropdown with ` Select the external object to which this object is related.`
![Uploading Screen Shot 2017-12-07 at 5.36.22 PM.png…]()

timrogers:
@LucasCioffi From what I remember, you just choose a type as normal (things like "Auto Number" are fine, for example) and it'll ask you if you want it to be an external ID at the next step.

[...]

PR #183, opened 2015-05-15
Add better documentation for #select

theSteveMitchell:
@ejholmes this should be a quick win.  In response to #182 I'm hoping to clarify the #select method.


timrogers:
@theSteveMitchell :+1: (I'll make some stylistic changes during my merge)

PR #185, opened 2015-06-03
Stop Restforce::Concerns::API#update! and Restforce::Concerns::API#upsert! mutating provided attributes

timrogers:
The use of `Hash#delete` in `Restforce::Concerns::API#update!` and `Restforce::Concerns::API#upsert!` means that the `attrs` passed into the Restforce client are mutated. It's generally not desirable for an external dependency like this gem to modify provided data in-memory.

This updates those methods to pull out the desired ID to update or upsert in a non-destructive way.


timrogers:
@ejholmes Thanks for all your great work on this gem! Should be a quick merge.

greysteil:
Would be great to get this merged - we're using this fork in production at the moment.

PR #186, opened 2015-06-10
Add a note about client object thread safety to the README

epbarger:
I had an issue trying to use Restforce with Sidekiq in a rails application. Turns out that reusing the client object causes a whole bunch of strange issues to manifest. I'm not positive if Restforce stores some state on the client or if it's an issue with Net::HTTP but I figured a note in the README would be nice to possibly help out other devs.


timrogers:
@epbarger :+1:

PR #189, opened 2015-06-28
Test against latest Ruby teeny versions

greysteil:
See https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/
for more details.

Alternative to https://github.com/ejholmes/restforce/pull/163.


timrogers:
:+1:

PR #190, opened 2015-06-29
Add #recent method to the client, reimplementing #150

timrogers:
New mergeable version of #150, with slightly improved specs and formatting.


timrogers:
FYI @davebrace

PR #198, opened 2015-08-17
Respect Faraday adapter option in authentication

stenlarsson:
There is another Faraday connection created in the authentication middleware, and this connection didn’t respect the adapter option, which meant that Net::HTTP was always used for that connection.


timrogers:
:+1: Merging this, despite the red on Travis which has broken in a bizarre case of bitrot.

timrogers:
I've just released this as part of v2.1.1 (<https://github.com/ejholmes/restforce/releases/tag/v2.1.1>) - it's on Rubygems too.

stenlarsson:
That was quick, thanks!

PR #210, opened 2015-10-07
Fix caching spec

kouno:
Faraday middleware has been updated and uses #read instead of #fetch.
MockCache has been updated to implement #read. #fetch remains in case we
want to support an older version of faraday_middleware.

See faraday_middleware commit:
https://github.com/lostisland/faraday_middleware/commit/3f559c18ee82701d8a1d04faf2bfd92d180b737b#diff-32d8ad0f070291407b5ceb8e788fb40fR47

Current failing travis build: https://travis-ci.org/ejholmes/restforce/jobs/81812544

### DIFF
--- a/lib/restforce/middleware/caching.rb
+++ b/lib/restforce/middleware/caching.rb
@@ -18,7 +18,7 @@ module Restforce
     end
 
     def use_cache?
-      !@options.key?(:use_cache) || @options[:use_cache]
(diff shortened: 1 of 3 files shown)


kouno:
@timrogers any comments on this?

timrogers:
@kouno Sorry for the delay! Looks good.

PR #212, opened 2015-10-12
Fix raise error parsing

### LINKED ISSUE #208: Parse body only if response header specify it
I'm using a custom APEX REST endpoint and when this endpoint validation fails, it returns a 500 and the body is a simple string.

It looks like Restforce is not checking the `Content-Type` header before parsing the body:

https://github.com/ejholmes/restforce/blob/66ef4acdd0517e7b31a4105af242c1948723157f/lib/restforce/middleware/raise_error.rb#L23

kouno:
This fixes #208. This branch is on top of `kouno:fix-cache-read` to see all tests pass.

### DIFF
--- a/lib/restforce/middleware/caching.rb
+++ b/lib/restforce/middleware/caching.rb
@@ -18,7 +18,7 @@ module Restforce
     end
 
     def use_cache?
-      !@options.key?(:use_cache) || @options[:use_cache]
+      @options.fetch(:use_cache, true)
     end
   end
 end
--- a/lib/restforce/middleware/raise_error.rb
+++ b/lib/restforce/middleware/raise_error.rb
@@ -16,11 +16,18 @@ module Restforce
     end
 
     def message
-      "#{body.first['errorCode']}: #{body.first['message']}"
+      "#{body['errorCode']}: #{body['message']}"
     end
 
     def body
-      JSON.parse(@env[:body])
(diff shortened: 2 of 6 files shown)


lib/restforce/middleware/caching.rb
```diff
@@ -18,7 +18,7 @@ def cache_key(env)
     end
 
     def use_cache?
-      !@options.key?(:use_cache) || @options[:use_cache]
+      @options.fetch(:use_cache, true)
```
timrogers: Good move here - if I'd written this I'd have use `#fetch` but didn't fix these when I did my refactor.

timrogers:
:+1:

timrogers:
I'll get this and #210 pushed as a new version ASAP.

kouno:
Thanks @timrogers!

lib/restforce/middleware/raise_error.rb
```diff
@@ -16,11 +16,18 @@ def on_complete(env)
     end
 
     def message
-      "#{body.first['errorCode']}: #{body.first['message']}"
+      "#{body['errorCode']}: #{body['message']}"
     end
 
     def body
-      JSON.parse(@env[:body])
+      @body = (@env[:body].respond_to?(:first) ? @env[:body].first : @env[:body])
```
kouno: This didn't entirely fix my issue as ActiveRecord actually monkey patch String to add the `first` method. :cry:

I'll add a PR to fix this again.

PR #217, opened 2015-11-16
Check for type instead of using duck typing

kouno:
When using ActiveSupport, the String class actually defines #first and
this would cause the body to be truncated to the first character.

The current tests are valid and still pass with this fix, but the
failing behavior only shows if you have active support included in your
project.

@timrogers sorry about that one. :see_no_evil:


timrogers:
:see_no_evil:

PR #220, opened 2015-12-08
Added authentication_callback documentation

anothermh:
I have documented the authentication_callback feature. This is an excellent feature that should be clearly explained given how finicky Salesforce can be with tokens.


timrogers:
Great job :+1:

PR #225, opened 2016-03-02
Raise a server error when authentication fails

greysteil:
Introduces a new error class, `ServerError`, which is raised when authentication fails with a 5xx code.


timrogers:
:+1:

PR #227, opened 2016-03-15
raise custom error if Salesforce responds with array.

michaelminter:
Fixes: https://github.com/ejholmes/restforce/issues/224

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -344,7 +344,11 @@ module Restforce
         response = api_patch "sobjects/#{sobject}/#{field}/#{external_id}",
                              attrs_without_field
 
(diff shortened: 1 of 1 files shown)


timrogers:
@michaelminter Looks good in principal. I'll have a bit of a play to see if there might be any nicer ways to do this, but will aim to get it merged in some form tomorrow.

timrogers:
@michaelminter I'm just looking at <https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm> and trying to work out exactly what the IDs returned are, and I can't quite work it out. Could you please send me an (appropriately anonymised) example response? Want to make the way the library handles this as understandable as possible, for example using the correct terminology.

timrogers:
Closing in favour of #228.

PR #228, opened 2016-03-16
Handle 300 errors triggered by an upsert where the external ID matches multiple records

### LINKED ISSUE #224: #upsert! TypeError: no implicit conversion of String into Integer
upsert! fails at lib/restforce/concerns/api.rb:347 with the above error message. Line in question:
```(response.body && response.body['id']) ? response.body['id'] : true```

```ruby
        external_id = attrs.
          fetch(attrs.keys.find { |k, v| k.to_s.downcase == field.to_s.downcase }, nil)
        attrs_without_field = attrs.
          reject { |k, v| k.to_s.downcase == field.to_s.downcase }
        response = api_patch "sobjects/#{sobject}/#{field}/#{external_id}",
                             attrs_without_field

        (response.body && response.body['id']) ? response.body['id'] : true
      end
```

Ruby 2.3.0 running on Heroku pointing to production salesforce, API Version 35.0. Passes fine running locally pointing to test.salesforce.com.

timrogers:
If you try to insert or update (upsert) a record using an external ID (see <https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm>) and the external ID provided is not unique, Salesforce returns a 300 response with a list of matching records as an array. This 300 behaviour can also be expected elsewhere in the API (<https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/errorcodes.htm>).

This allows us to handle that response and raise an exception, as you'd expect, and provides access to the list of matching IDs returned, since once can access the response body of a `Faraday::ClientError` with `#response` (<http://www.rubydoc.info/github/lostisland/faraday/Faraday/ClientError>).

Fixes #224.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -315,7 +315,8 @@ module Restforce
       #
       # Returns true if the record was found and updated.
       # Returns the Id of the newly created record if the record was created.
-      # Returns false if something bad happens.
+      # Returns false if something bad happens (for example if the external ID matches
+      # multiple resources).
       def upsert(*args)
         upsert!(*args)
       rescue *exceptions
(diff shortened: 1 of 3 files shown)


timrogers:
@greysteil It'd be great if you could take a quick look at this.

greysteil:
:+1:

PR #232, opened 2016-04-07
encode external id for upsert! to get rid of error with multibytes external id

nagachika:
When I pass strings contains multibytes characters as external ID to Restforce::Concerns::API#upsert!, it raise `URI::InvalidURIError:URI must be ascii only` exception.
To get rid the failure, URI encode the external ID when embeded it into request path.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -395,7 +395,7 @@ module Restforce
       #
       # Returns the Restforce::SObject sobject record.
       def find(sobject, id, field = nil)
-        url = field ? "sobjects/#{sobject}/#{field}/#{id}" : "sobjects/#{sobject}/#{id}"
+        url = field ? "sobjects/#{sobject}/#{field}/#{URI.encode(id)}" : "sobjects/#{sobject}/#{id}"
         api_get(url).body
       end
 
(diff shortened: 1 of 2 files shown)


timrogers:
@nagachika Sorry for the slow response! Can object or field names contain multibyte characters? If so, we might want to look at a more holistic solution to this issue.

timrogers:
@nagachika *ping* 🎾

nagachika:
Sorry for a long absence.

> Can object or field names contain multibyte characters?

I don't think so. `external_id` is the value of object but `object` and `field` are derived from API reference name and contains only ASCII characters.

lib/restforce/concerns/api.rb
```diff
@@ -344,7 +345,7 @@ def upsert!(sobject, field, attrs)
           fetch(attrs.keys.find { |k, v| k.to_s.downcase == field.to_s.downcase }, nil)
         attrs_without_field = attrs.
           reject { |k, v| k.to_s.downcase == field.to_s.downcase }
-        response = api_patch "sobjects/#{sobject}/#{field}/#{external_id}",
+        response = api_patch "sobjects/#{sobject}/#{field}/#{URI.encode(external_id)}",
```
timrogers: This looks sensible to me. I had a look through how encoding works, and elsewhere Faraday will do the encoding automatically.

@nagachika Can you add the same to the `#find` and `#select` methods (which also use an external ID)? Then I'll merge this in.

nagachika: @timrogers 
Good catch! I've added patch for `#find` and `#select` methods.

nagachika:
I pushed some code cleaning to pass robocop check.

PR #245, opened 2016-07-12
set SF api version from env var

jhelbig:
All other options have a related ENV var from the system that's hosting this gem.  api_version defaults to 26.0 but there's no way to set this from an ENV var upon deployment.  This change will allow the variable to be set or it will default to v26.0 if none was specified.

### DIFF
--- a/lib/restforce/config.rb
+++ b/lib/restforce/config.rb
@@ -88,7 +88,7 @@ module Restforce
       end
     end
 
-    option :api_version, default: '26.0'
+    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '26.0' }
 
     # The username to use during login.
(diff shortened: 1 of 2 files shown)


timrogers:
Great idea! Can you add a spec? (P.S. I can see that tests are failed due a problem installing Rack - I'll look into it ASAP!)

timrogers:
If you pull down and then `git rebase master`, that should get the specs working 🚀

jhelbig:
Thanks for the quick response.  I have merged everything into the patch-1 and master branches and all tests are looking good.

timrogers:
:+1:

timrogers:
This has been released in https://github.com/ejholmes/restforce/releases/tag/v2.3.0.

jhelbig:
thanks again for the quick commits!

PR #251, opened 2016-07-22
Add Salesforce Document support

jhelbig:
Noticed documents were not supported.  Attachments and Documents are similar enough and both contain 'Body' attributes that host the file information.  This was essentially a copy and replace with the Attachment class.

### DIFF
--- a/lib/restforce/mash.rb
+++ b/lib/restforce/mash.rb
@@ -24,7 +24,7 @@ module Restforce
           # of sobject records.
           Restforce::Collection
         elsif val.key? 'attributes'
-          case(val['attributes']['type'])
+          case (val['attributes']['type'])
           when /^Attachment$/
             Restforce::Attachment
           when /^Document$/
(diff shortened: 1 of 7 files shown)


timrogers:
Still meaning to do this - will review it as soon as I can 👼

jhelbig:
no worries, the project I'm working on can use the fork until you're ready to review.

Thanks for the update!

lib/restforce/mash.rb
```diff
@@ -24,8 +24,11 @@ def klass(val)
           # of sobject records.
           Restforce::Collection
         elsif val.key? 'attributes'
-          if val['attributes']['type'] == 'Attachment'
+          case (val['attributes']['type'])
+          when /^Attachment$/
```
timrogers: A bit odd (and probably some overhead) to use a regular expression here, how about

```ruby
when "Attachment"
  ...
```

timrogers:
@jhelbig Ready to go apart from the one comment! :rocket:

jhelbig:
agreed.  You get stuck doing regex for complicated comparisons it's hard to not use it when unnecessary.

PR #259, opened 2016-10-12
databasedotcom deprecated. remove from readme

espen:
From [databasedotcom readme](https://github.com/heroku/databasedotcom): "DEPRECATED: Please use restforce instead."


timrogers:
👍

Issue #260, opened 2016-10-17
not compatible with json 2.0 (and thus ruby 2.4)


ccutrer:
JSON 2.0.0 dropped support for ruby < 2.0.0, but is otherwise compatible. JSON 2.0.0 is required to be compatible with Ruby 2.4. Thus until restforce updates its dependency constraint, restforce is not compatible with ruby 2.4

timrogers:
@ccutrer Good spot! Have you tried with JSON 2.0.0? What's broken?

ccutrer:
I haven't yet. I'm working on just trying to run our app with Ruby 2.4, and I have several other dependency problems to work through first. restforce is an optional feature in our app, so I just moved that code out for now. AFAICT, though, nothing should be broken - the only reason for the major version bump was dropping of ruby 1.8/1.9 support. Which you don't even need to worry about, as long as your dependency requirement still allows json < 2.0, then the consumers of your gem can pick the version of json that will work with the version of ruby they're using.

timrogers:
I'll have a play now and see if it works - I imagine it will...

timrogers:
I've put together a fix to this in https://github.com/ejholmes/restforce/pull/261 - looks like it's behaving well with the tests passing on all versions.

PR #261, opened 2016-10-18
Relax json dependency to support Ruby 2.4.0-dev

timrogers:


timrogers:
Fixes #260

restforce.gemspec
```diff
@@ -20,7 +20,15 @@ Gem::Specification.new do |gem|
...
   gem.add_dependency 'faraday', '~> 0.9.0'
   gem.add_dependency 'faraday_middleware', '>= 0.8.8'
-  gem.add_dependency 'json', ['>= 1.7.5', '< 1.9.0']
+
+  # Ruby 2.4 requires JSON 2.0, but that doesn't work with old pre-2.0.0 versions of Ruby.
+  # See https://github.com/ejholmes/restforce/issues/260.
+  if RUBY_VERSION =~ /^1.9/
+    gem.add_dependency 'json', ['>= 1.7.5', '< 1.9.0']
+  else
+    gem.add_dependency 'json', '>= 2.0.0'
```
ccutrer: this is a bit draconian. now if you're running say ruby 2.3, you _have_ to update json to 2.0.0 - which can be difficult (read: impossible) if you have other gems with the same problem of not allowing json 2.0

ccutrer:
Yup, that would work fine. Then if anyone wants to use ruby 2.4, it's their job to make sure they're using at least 2.0.

Issue #237, opened 2016-05-17
Replay PushTopic Streaming Events (Beta)


kris-luminar:
SalesForce has a new feature for PushTopics that will allow replaying events that where missed by clients.
https://developer.salesforce.com/docs/atlas.en-us.api_streaming.meta/api_streaming/replay_pushtopic_events.htm

My team thinks that this feature only exists in version 36 of SalesForce's API. 

This feature would be very helpful for my team. Are there any plans to support this new functionality?

anacaet:
Someone has an updates on this topic?

panozzaj:
I implemented something that might work for this, I invite you to check out PR #265 and leave any feedback or comments.

PR #265, opened 2016-10-21
Implement push topic replay

panozzaj:
Thanks to the existing code in Restforce, I was able to get up and running with the PushTopic streaming API in just a few lines of code. We have used this library on a couple of projects now.

For PushTopics, Salesforce added [the ability to replay messages](https://developer.salesforce.com/docs/atlas.en-us.api_streaming.meta/api_streaming/using_streaming_api_durability.htm) (see #237). This PR adds an options hash to the `subscribe` method on clients, which allows us to pass a message id to replay from using a client extension. With this code, you can replay messages:

```
client.subscribe push_topic_name, replay: 7 do |message|
  # message handler here
end
```

This will replay message 8 and up (based on Salesforce's documentation of this feature.) I have verified the functionality with a local script that points to this PR's version of the gem.

---

I am pushing this up for initial feedback. I think it needs at least a little more work.

One thing is that the current design does not let you easily subscribe to some channels with a replay id, and subsequently subscribe to others with no replay id. This is because there is only one Faye client initialized in total (it is cached between `subscribe`s.)

I think we might be able to remove the extension by storing it and then removing it later if we subscribe with no replay id, but it seems a little clunky. I wanted to see what the intent of the `faye` method is or why we want to only create one Faye client in the existing code. It probably didn't matter before since we were never modifying the client, but now that we are, it is a little more complicated. I am guessing the method was made public just for testing purposes based on the associated test and a search of the code, but wasn't sure of the intent of it. It seems like if we made a new client for each call to `subscribe` that we could easily pass in replay id or not and not have any interference between calls.

Additionally, I'd like to add tests and documentation for this feature and clean up the the logging (it's a bit verbose.) I'd also be curious to know if there is a more preferred place to put the `ReplayExtension` class, since I just lumped it in to the existing streaming module file.

---

Note that if you try to replay a message that is already expired (was created 24 hours or more ago), you will get the following error when attempting to subscribe to the channel:

```
#<Faye::Error:0x007fe9361df548 @code=nil, @params=[], @message="400::The replayId {2} you provided was invalid.  Please provide a valid ID, -2 to replay all events, or -1 to replay only new events.">
```

The best solution would seem to be to either increase the number you are replaying until you find a good one, or to use "-2" to replay all possible messages that Salesforce still has.

---

Useful resources:
- [cometdReplayExtension.js](https://github.com/developerforce/StreamingReplayClientExtensions/blob/48ce54bb9ef939ff13c73302a0f33786c5c343f2/javascript/cometdReplayExtension.js) (this gave a lot of detail that was useful)
- [Faye client extensions documentation](https://faye.jcoglan.com/ruby/extensions.html) (pretty much used this exact template for the extension)
- [Faye Ruby client example](https://github.com/faye/faye/blob/master/examples/ruby/client.rb)

### DIFF
--- a/lib/restforce/concerns/streaming.rb
+++ b/lib/restforce/concerns/streaming.rb
@@ -44,12 +44,12 @@ module Restforce
       end
 
       class ReplayExtension
-        def initialize options
+        def initialize(options)
           @channels = options[:channels]
           @replay = options[:replay]
         end
 
-        def outgoing message, callback
+        def outgoing(message, callback)
           # Leave non-subscribe messages alone
           unless message['channel'] == '/meta/subscribe'
             return callback.call(message)
--- a/lib/restforce/concerns/streaming.rb
+++ b/lib/restforce/concerns/streaming.rb
@@ -91,7 +91,6 @@ module Restforce
             handler
           end
         end
-
       end
     end
   end
--- a/lib/restforce/concerns/streaming.rb
+++ b/lib/restforce/concerns/streaming.rb
@@ -7,8 +7,10 @@ module Restforce
       # block    - A block to run when a new message is received.
       #
       # Returns a Faye::Subscription
-      def subscribe(channels, &block)
-        faye.subscribe Array(channels).map { |channel| "/topic/#{channel}" }, &block
+      def subscribe(channels, options = {}, &block)
+        @replay = options[:replay]
+        @channels = Array(channels)
+        faye.subscribe @channels.map { |channel| "/topic/#{channel}" }, &block
       end
 
       # Public: Faye client to use for subscribing to PushTopics
@@ -19,17 +21,49 @@ module Restforce
 
         url = "#{options[:instance_url]}/cometd/#{options[:api_version]}"
 
-        @faye ||= Faye::Client.new(url).tap do |client|
-          client.set_header 'Authorization', "OAuth #{options[:oauth_token]}"
+        unless @faye
+          @faye = Faye::Client.new(url).tap do |client|
+            client.set_header 'Authorization', "OAuth #{options[:oauth_token]}"
 
-          client.bind 'transport:down' do
-            Restforce.log "[COMETD DOWN]"
-            client.set_header 'Authorization', "OAuth #{authenticate!.access_token}"
+            client.bind 'transport:down' do
+              Restforce.log "[COMETD DOWN]"
+              client.set_header 'Authorization', "OAuth #{authenticate!.access_token}"
+            end
+
+            client.bind 'transport:up' do
+              Restforce.log "[COMETD UP]"
+            end
           end
 
-          client.bind 'transport:up' do
-            Restforce.log "[COMETD UP]"
+          if @replay
+            @faye.add_extension ReplayExtension.new(channels: @channels, replay: @replay)
           end
+        end
+
+        @faye
+      end
+
+      class ReplayExtension
+        def initialize options
+          @channels = options[:channels]
+          @replay = options[:replay]
+        end
+
+        def outgoing message, callback
+          # Leave non-subscribe messages alone
+          unless message['channel'] == '/meta/subscribe'
+            return callback.call(message)
+          end
+
+          # Set the replay value for the each channel
+          message['ext'] ||= {}
+          message['ext']['replay'] = {}
+          @channels.each do |channel|
+            message['ext']['replay']["/topic/#{channel}"] = @replay
+          end
+
+          # Carry on and send the message to the server
+          callback.call message
         end
       end
     end
--- a/spec/unit/concerns/streaming_spec.rb
+++ b/spec/unit/concerns/streaming_spec.rb
@@ -17,7 +17,6 @@ describe Restforce::Concerns::Streaming, event_machine: true do
     end
 
     context "replay_handlers" do
-
       before {
         faye_double.should_receive(:subscribe).at_least(1)
         client.stub faye: faye_double
@@ -42,7 +41,6 @@ describe Restforce::Concerns::Streaming, event_machine: true do
           'channel3' => 3
         )
       end
-
     end
   end
 
@@ -69,7 +67,8 @@ describe Restforce::Concerns::Streaming, event_machine: true do
         faye_double.should_receive(:set_header).with('Authorization', 'OAuth secret2')
         faye_double.should_receive(:bind).with('transport:down').and_yield
         faye_double.should_receive(:bind).with('transport:up').and_yield
-        faye_double.should_receive(:add_extension).with(kind_of(Restforce::Concerns::Streaming::ReplayExtension))
+        faye_double.should_receive(:add_extension).with \
+          kind_of(Restforce::Concerns::Streaming::ReplayExtension)
         subject
       end
     end
@@ -79,7 +78,6 @@ describe Restforce::Concerns::Streaming, event_machine: true do
         expect { subject }.to raise_error
       end
     end
-
   end
 
   describe Restforce::Concerns::Streaming::ReplayExtension do
@@ -129,7 +127,7 @@ describe Restforce::Concerns::Streaming, event_machine: true do
         }
       }
 
-      extension.incoming(message, -> m { })
+      extension.incoming(message, -> m {})
       handler.should eq('channel1' => 42)
     end
 
@@ -142,7 +140,7 @@ describe Restforce::Concerns::Streaming, event_machine: true do
         'data' => {}
       }
 
-      extension.incoming(message, -> m { })
+      extension.incoming(message, -> m {})
       handler.should eq('channel1' => 41)
     end
 
@@ -163,6 +161,5 @@ describe Restforce::Concerns::Streaming, event_machine: true do
     def read_replay(message)
       message.fetch('ext', {})['replay']
     end
-
   end
 end
--- a/spec/unit/concerns/streaming_spec.rb
+++ b/spec/unit/concerns/streaming_spec.rb
@@ -146,11 +146,11 @@ describe Restforce::Concerns::Streaming, event_machine: true do
 
     private
 
-    def subscribe(extension, to:)
+    def subscribe(extension, options = {})
       output = nil
       message = {
         'channel' => '/meta/subscribe',
-        'subscription' => "/topic/#{to}"
+        'subscription' => "/topic/#{options[:to]}"
       }
       extension.outgoing(message, -> m {
         output = m


georges:
+1

panozzaj:
Thanks for the vote of confidence! I forgot that I pushed up a repo that demonstrated the replay ability at https://github.com/OurHealth/pushtopic-listener. Hope this helps!

timrogers:
I'm not against this but I don't have any experience whatsoever of the realtime stuff in Restforce or the Salesforce API. Sorry for my slowness on this one @panozzaj!

@georges Would you fancy giving this a thorough review?

panozzaj:
Cool, thanks for the response @timrogers! Yeah the PushTopic stuff seems different from the rest of the code, and it was pretty fun to dive into the internals and learn more about Faye. The code has worked pretty well so far based on local testing. I still think it needs documentation / tests, etc. but think it is a step in the right direction. Thanks again!

btgtech:
@panozzaj I'm no expert in Faye extension but I've used this patch in production for several weeks now and it's working well. I used the -2 replay all parameter and seems to work as advertised every time our listener restarts, which is often due to continuous deployments.

georges:
@timrogers I'm no expert in Faye extension but I've used this patch in production for several weeks now and it's working well. I used the -2 replay all parameter and seems to work as advertised every time our listener restarts, which is often due to continuous deployments.

aliibrahim:
Any ideas when will this branch merged into master?

timrogers:
@aliibrahim Sorry for the wait on reviewing this! I'll aim to take a look in the next few days so we can get it merged. Seems promising given that some above have been running it successfully.

@georges Is that still the case?

btgtech:
@timrogers There is one problem with the patch. If you want to use a replayId to represent that last known event you're processed, the replayId that is passed initially in the block is always used in case the connection goes down and back up. We've experienced a lot of disconnects. This means that every time the listener connects back to Salesforce, it will pass the replayId that was original setup in the block, which means that you will end up replaying an unnecessary amount of past events.
There needs to be a mechanism for the gem to request a current replayId. Perhaps use the return value for the block that's being yielded? Hope that makes sense.

timrogers:
@btgtech That sounds like a bit of a blocker to merging this. Your block suggestion makes sense to me, but is it doable with how Faye works? (I'm not familiar and there's no direct `yield` here we can work with.)

panozzaj:
OK, thanks for that feedback! I think that it is worth exploring to see if there is a way to pass a different id. This actually might be related to the comment I made in the description:

> One thing is that the current design does not let you easily subscribe to some channels with a replay id, and subsequently subscribe to others with no replay id. This is because there is only one Faye client initialized in total (it is cached between subscribes.)

I have seen disconnects on our end as well. I am not sure how to detect when they happen though. Do you hook into a Faye lifecycle event of some sort? Wondering partially so we can better handle reconnects on our end, and partially because it will help me understand how I might implement something like this. Thanks!

panozzaj:
> I have seen disconnects on our end as well. I am not sure how to detect when they happen though. Do you hook into a Faye lifecycle event of some sort? Wondering partially so we can better handle reconnects on our end, and partially because it will help me understand how I might implement something like this. Thanks!

@btgtech can you provide more context for how you are detecting when the disconnect happens?

panozzaj:
@btgtech @timrogers: @drteeth made [a PR](https://github.com/OurHealth/restforce/issues/1) against this branch. I believe that it addresses the concern you had about passing the original `replayId` on restarts. Also, now we have a separate `replayId` per channel and are able to remember what the last seen `replayId` was for each channel. What do you think about this approach? Thanks!

lib/restforce/concerns/streaming.rb
```diff
@@ -30,6 +31,65 @@ def faye
...
+      end
+
+      class ReplayExtension
+        def initialize(replay_handlers)
+          @replay_handlers = replay_handlers
+        end
+
+        def incoming(message, callback)
+          channel = message.fetch('channel').gsub('/topic/', '')
+          replay_id = message.fetch('data', {}).fetch('event', {})['replayId']
```
georges: Seems to me that saving the replay_id should occur after the callback successfully return

drteeth: Good idea. I'll make the change soon.

georges:
There is one problem related to how SF handles replay ids. What I've observed is that if you subscribe with a replayId that is out of SF retention window, SF will simply stop sending any events, including *newer* events.

This is an issue in case the connection drops and needs to be reconnected. If the last replayId that was stored is for an event that's older than 24, the extension will re-subscribe to the channel with that id and it will cause SF to stop sending anything.

Asked the question there but got nowhere

https://developer.salesforce.com/forums/?id=9060G000000I3rlQAC

I think what the extension needs to do is that in addition to the replayId, it needs to store the timestamp for the event. If the replayId correspond to an event that's older than 24, it should not use that replayId to subscribe.

drteeth:
By 24, you mean 24 hours I assume?

If I understand you correctly, you are suggesting we should store the time at which we store the replayId, and if when we read it back it is older than the 24 hour retention window, we should ignore it and return the default instead.

I see the issue, and I'm hesitant to push that onto client's of the library because it's subtle. It's unfortunate because as it stands the interface to the replay handler is essentially a hash, and was intended to work with hashes.

Maybe we can see if the handler responds to a method to store the timestamp and only conditionally use that?

LMK if I've misunderstood the issue.

georges:
yes, sorry meant 24 hours. WRT to timestamp, there is a 'createdDate' in the event sent by SF, which I'd assume is what they use to determine retention window. so yes, basically we can't use the event id as replayId for an event that has 'createdDate' older than 24 hours ago otherwise SF 'freezes'.

I don't have a strong opinion on implementation details.

lib/restforce/concerns/streaming.rb
```diff
@@ -30,6 +31,65 @@ def faye
...
+          @replay_handlers = replay_handlers
+        end
+
+        def incoming(message, callback)
+          callback.call(message).tap do
+            channel = message.fetch('channel').gsub('/topic/', '')
+            replay_id = message.fetch('data', {}).fetch('event', {})['replayId']
+
+            handler = @replay_handlers[channel]
+            if !replay_id.nil? && !handler.nil? && handler.respond_to?(:[]=)
```
georges: I think there is a bug here. If the replay_id is initialized with a scalar value, handler.respond_to?(:[]=) will be false and the replay_id is never updated. The "remember last replay_id" assignment logic should probably mirror what the replay_id(channel) method is doing.

drteeth: @georges I see what you mean. 

When I wrote it, I had in mind that if you pass a scalar it will not be updated ever. The idea being that I want to specify one of the magic values for this subscription:

* -1  Subscriber receives new events that are broadcast after the client subscribes.
* -2  Subscriber receives all events, including past events that are within the 24-hour retention window and new events sent after subscription.

It was my intent to have hashes cover the case where you want to track the latest values. Because we're mutating the replay value, it can't be a scalar.

I admit that that isn't immediately obvious, and I'm not altogether happy with the reference semantics involved with this implementation anyway. Meaning I don't really like that you pass an object to the subscription that you expect it to mutate for you.

All that said I think this could work as is, even if it's not ideal. perhaps we can replace -1 and -2 with :new_events and :past_events or similar?

Thoughts?

drteeth:
I envisioned 3 use cases for the reply IDs:

Magic value Scalars (-1 and -2): on (re)subscription always ask for all the events in the current window. The intent is that the consumer of the messages will handle previously seen events
```ruby
client.subscribe("foo", replay: -2)         # always send -2
```

Hashes: on subscription start with the given ID. Each received message will update the replay Id. On re-subscription, the latest replay Id will be used and the client will pickup where it left off. Clearly the values are only stored in memory here and it would be up to user to make them durable.

```ruby
replay_ids = {foo: -2}
client.subscribe("foo", replay: replay_ids)  # Start with -2 and update with each message
```

Custom: any time a replay ID is needed or received, the indexer methods are called. The intent is to allow custom handling of the replay IDs. Perhaps they are written off to a database or published on another event stream.

```ruby
class ReplayHandler
  @datastore = ...

  def []=(key, value)
    @datastore.write(key, value)
  end

  def [](key)
    @datastore.read(key)
  end
end

custom_handler = ReplayHandler.new
# the handler will determine the starting Id and how to 
client.subscribe("foo", replay: custom_handler)  remember subsequence values.
```

lib/restforce/concerns/streaming.rb
georges: That make sense but wasn't obvious just looking at method signature. Mapping magic values to symbol might indeed make the behavior more clear.

georges:
@drteeth ah, i had missed the support for custom replay handler. that case makes a lot of sense to offer the flexibility needed to persist replay id.

panozzaj:
Sorry, I had not followed this thread much (we ended up going with a different solution for listening to PushTopics at work). Just wanted to see if I could try to summarize or see where this PR stands. Are the following still an issue and/or the only outstanding issues?

 - "we can't use the event id as replayId for an event that has 'createdDate' older than 24 hours ago otherwise SF 'freezes'." -- there has been some discussion of using the custom handler for this, not sure where this stands. I think at least the custom handler functionality has been implemented.
 - Changing from magic scalars (`-1`, `-2`) to something more semantic. If this was an issue, could this be a separate pull request since it could be considered an enhancement?

I am wondering if we need README-level / documentation changes since this is an additional feature. It could just be a slight expansion of [this comment](https://github.com/ejholmes/restforce/pull/265#issuecomment-300581597) since it has some good examples, and otherwise mention this PR's intent.

georges:
@panozzaj agreed. custom handler can be used to solve for replayId staleness, and more semantic -1,-2 is a nice enhancement, not a show stopper.

yaronyagoda:
@panozzaj 
Hi,
I am trying this branch to consume Push topics
I have managed to create a listener but I do see disconnections after a while, how should I catch and handle them?
Thanks
Yaron

neilmiddleton:
This PR looks good to me.  I'm not fussed about the `-1, -2` stuff but otherwise I'd love to see this merged.

yaronyagoda:
@neilmiddleton Hi Neil
Are you using Restfoce is a push topic listener?
If so, do you experience connection getting lost?
Thanks

drteeth:
@yaronyagoda I have not experienced the connection getting lost, what happens exactly? The intent is that it should resume on re-connection with the last replayId that it saw if you are using a hash or a custom replay. Otherwise, it should just pass -1 or -2 upon reconnection.

neilmiddleton:
@yaronyagoda We're not running it in production yet, but tests locally suggest that there's a certain amount of reconnection already built in to Faye.  Aside from that I've just added some logic locally to pass in a replayId should one be present and that seems to work fine.

Ideally, if connection is lost, an exception should be raised by RestForce so you can attempt a new subscription, but I don't see that within the scope of this PR>

yaronyagoda:
@neilmiddleton 
Thanks for the quick reply
I am also not using it in production.
We are using JS code using jsforce to listen to push topics which is not reliable.
I managed to have a listener (using a rake task) that listens to a specific topic.
But sometimes after a few hours, I stop getting messages...
I haven't tested it enough but I believe it does get reconnect but I am not sure.
I would like to be notified upon this disconnect so I can subscribe again but I do not see any event/exception

zrc210:
@yaronyagoda @neilmiddleton While not related to the replay functionality in this PR, I am experiencing silent disconnects as well, as detailed in https://github.com/ejholmes/restforce/issues/318

Unfortunately, I am running in production, and the dirty solution has been to restart the script that subscribes to the push topics on a daily basis.

panozzaj:
We had also experienced some disconnects (at least on an earlier version of this PR), as had @btgtech (at least based on https://github.com/ejholmes/restforce/pull/265#issuecomment-276154147). I don't have a good solution for detecting and automatically reconnecting at this point. I like the analysis/proposal in #318.

I guess the question in my mind would be: should we wait to merge this PR until we figure out the disconnect/reconnect story because it would generate lots of questions/issues, or does this move the ball forward since people are apparently using it in production to some degree? Maybe a caveat in the docs or more documentation of how people are reconnecting would help?

zrc210:
@panozzaj FWIW - I feel that this PR is indirectly related to the disconnect issue that has been going undetected for awhile.

I don't think solving or diagnosing the disconnect issue is really within scope of this PR, and the decision as to when this PR should be merged should not depend on the disconnect issue.

yaronyagoda:
@panozzaj @zrc210 
I have great news!!!
I managed to find the problem for the disconnects
It's a bug on Faye that they don't listen to disconnect messages
I want to write a PR here next week and I would like a review.
In the meantime I override it in my code:

I have added an extension to Faye:

class ReplayExtension
     def incoming(message, callback)
    if message['channel'] == "/meta/disconnect"
       Rails.logger.info("ReplayExtension - restarting connection")
        # do your logic to restart the server
      end
      callback.call(message)
    end
 
     def outgoing(message, callback)
end
end

And indeed I do an auto restart every few hours

If you want to see the faye logs (this is how I found it )
Faye.logger = lambda { |m| puts m }
 Faye::Logging.log_level = :debug

And after googling it I saw that this how the guys at NForce fixed (a js library working with Faye)

So good new!!!

neilmiddleton:
Regarding disconnects, in Version 28.0 API document (Summer '13), Salesforce state as below (which is not placed in later docs);

> Maximum CometD session lifetime: 2 hours
A client subscription is valid for a maximum of 2 hours, irrespective of activity on the connection. Therefore the client must handshake, subscribe, and connect before the session is timed out on the server side. A new subscription creates a new server session: the server timeout clock restarts again.

Therefore you need to resub every 2 hours.  I'm handling it like this:

```ruby
def listen_salesforce
  EM.run do
    connect_to_salesforce

    EM.add_periodic_timer(1800) do
      salesforce_client.faye.unsubscribe(["/topic/MyTopic"])
      connect_to_salesforce
    end
  end
end

def connect_to_salesforce
  salesforce_client.authenticate!
  salesforce_client.subscribe 'MyTopic' do |message|
      receive_salesforce_message(message)
    end
end
```

This causes a complete reauth and handshake every 30 minutes and seems, from some early testing, to work fine.

yaronyagoda:
@panozzaj - do you have any plans to merge this branch to master? 
Thanks
Yaron

panozzaj:
Thanks Yaron! To my knowledge I do not have merge rights to this repo. I don't think there are any identified blockers to merge, I think the ball is in the maintainers' court. Would get a +1 from me.

timrogers:
In that case, I'll try to take a look at this as soon as I can. What testing have you done @panozzaj?

panozzaj:
Greetings! Based on the above discussion I believe that the ball is in my court. However, I recently left the job that I was working on this with, so I don't have a viable Salesforce test environment. So I will have to leave it to you all to do any final testing. I believe others have tested this in some manner. Sorry to leave you hanging, hope this helps move the PR forward.

ivanovyordan:
Hi guys. I just tested the changes and the replays works perfectly. Can I help somehow and move the PR forward?

drteeth:
I think what is missing is updates to the docs describing how to use the 3 flavours of replay handlers: scalar, hash and custom.

panozzaj:
Makes sense, we could build off [your excellent comment](https://github.com/restforce/restforce/pull/265#issuecomment-300581597).

timrogers:
I would love to get this released 😍

pre:
@timrogers Are there any blockers getting this merged? :)

timrogers:
From my perspective, I think we just need some documentation now.

pre:
Could someone, who is already invested in this PR, write a minimal amount of documentation so that we'd get this in master? Ping @panozzaj @drteeth @ivanovyordan  @btgtech 

@timrogers Would it be enough to document the following?
* "Example of a custom replay handler with `ReplayExtension`" 
* Consider persisting the last known good replay_id somewhere where you can access it when the client connects the next time.
*  Magic replay_ids are `-1` to replay only new events arriving after the client connects and `-2` to replay all events since last 24 hours.
* Client **must** provide a `replay_id` which is for a message younger than 24 hours. A replay id for a message older than 24 hours will cause Salesforce to not send anything, not even newer events.

timrogers:
Yes, that sounds good to me.

Usually, I'd offer to do it myself, but I'm not really familiar enough with
the functionality in this case.

pre:
@andreimaxim Above the TODO list for you to be the hero of the day!

pre:
@andreimaxim How about having a celebratory summer push for this ? :)

andreimaxim:
@pre sorry about the delay, I thought I had some time to review the code and write the documentation but I had to move away from Salesforce-land for a bit. Will do my best to have it ready for Monday ;-)

andreimaxim:
@pre I've added a PR to @panozzaj 's feature branch which includes documentation for the push topic replay feature. Once it's merged there the commit should also appear in this PR.

panozzaj:
I think the documentation @andreimaxim crafted looks fantastic!

Current blocker is that I changed jobs so I no longer have commit access to the repo that the feature branch was in, so I can't merge it... 🥇

Few options at this point that I can see:

1. \[*My preferred option*\] If this is the only thing holding these changes up, I think the easiest path would be for one of the maintainers to merge this PR (#265) and then get the documentation at  https://github.com/OurHealth/restforce/pull/3 by cherry-picking it over or adding a new PR here for it.
2. Alternately, I could fork https://github.com/OurHealth/restforce, open a new PR on https://github.com/restforce/restforce which points to this one, and we close this PR and then merge the new one. It might be harder to trace back through history, but wouldn't be that much work.

I haven't run into this particular problem before, so maybe there is there some easy Github way that I do not know about to merge the documentation PR in or switch which feature branch this PR is based off of.

andreimaxim:
+1 for option 1.

pre:
Great job @andreimaxim ! 

@timrogers could you help us to merge this (see the two comments above) ?

timrogers:
Sorry for my slow work on this! I'll get it reviewed in the next day or two, and hopefully finally merged! :+1:

pre:
@timrogers GREAT! Thanks!

PR #266, opened 2016-10-31
bump default api_version to 38.0

thomasdziedzic:
fixes issues with older versions, for example: https://github.com/ejholmes/restforce/issues/167 describes one such potential issue with older api versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ end
 
 ### API versions
 
-By default, the gem defaults to using version 26.0 (Winter '13) of the Salesforce API.
+By default, the gem defaults to using version 38.0 (Winter '17) of the Salesforce API.
 Some more recent API endpoints will not be available without moving to a more recent
 version - if you're trying to use a method that is unavailable with your API version,
 Restforce will raise an `APIVersionError`.
--- a/lib/restforce/config.rb
+++ b/lib/restforce/config.rb
@@ -88,7 +88,7 @@ module Restforce
       end
     end
 
-    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '26.0' }
+    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '38.0' }
 
     # The username to use during login.
     option :username, default: lambda { ENV['SALESFORCE_USERNAME'] }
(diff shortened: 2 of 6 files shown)


timrogers:
Hi @thomasdziedzic! I'm not totally against this, but I suspect we'd need a major version for the library as upping the version will almost certainly lead to backwards-incompatible changes.

Could we consider just changing the README to point people to set the version to a newer one?

thomasdziedzic:
Could you give me an example of a backwards incompatible change that could happen?
I always thought that Salesforce tries to keep backwards compatibility as much as possible.

IMO the version bump is necessary to keep the defaults for new users predictable. We could do a major version bump for the gem to reflect this change, but at the end of the day, I think it would be less error prone to use the latest version of the api as the default for the gem for users.

timrogers:
I'll look into it - I assumed that Salesforce API versions would be likely to introduce breaking changes but I could be wrong! 

If Salesforce API versions introduce breaking changes, then we can't really change this default without a major version (as we want to stick to semantic versioning) which seems a bit overkill to me - otherwise, a minor version update could break people's code. Leave it with me and I'll do some research.

timrogers:
I've just done some more research on this and the basic answer is that we won't be able to change this without a major version release (e.g. `3.0.0`)

Between API versions, the API can change significantly (e.g. see [this Quora answer](https://www.quora.com/Is-it-true-that-no-application-written-on-SalesForce-has-ever-been-rendered-obsolete-by-an-update-every-app-every-written-for-SalesForce-is-still-operable-How-does-SF-ensure-such-complete-backward-compatibility) which shows differences in the Chatter users API between versions 23 and 32 - fields have changed completely). This would make any change backwards-incompatible, potentially breaking people's code unless they (a) make changes to their implementation or (b) manually specify the earlier version they want. 

Salesforce keep their old API versions around forever, so I feel like there's no pressing need to force the issue. I'll look at making this change when there's a good reason to go to `3.0.0`, as it feels a bit silly to release a major version just to up the API version.

For now, I'll update the README to suggest that people specify the latest version - this will get the best value and the newest features for users.

Thanks for your time on this issue - it's definitely a good thing to change! :+1:

PR #273, opened 2016-11-28
No credentials in cache keys

atmos:
While instrumenting our code with zipkin we noticed the auth headers being included in cache keys. While hashing them isn't perfect it's better than what we're doing now.

### DIFF
--- a/Gemfile.travis
+++ b/Gemfile.travis
@@ -4,3 +4,4 @@ gemspec
 gem 'rake'
 gem 'rack', '~> 1.6.4'
 gem 'jruby-openssl', :platforms => :jruby
+gem 'public_suffix', '2.0.3'
--- a/lib/restforce/concerns/caching.rb
+++ b/lib/restforce/concerns/caching.rb
@@ -19,6 +19,10 @@ module Restforce
       def cache
         options[:cache]
       end
+
+      def params_to_ignore
+        (super + ["Auth"]).uniq
+      end
     end
   end
 end
--- a/lib/restforce/concerns/caching.rb
+++ b/lib/restforce/concerns/caching.rb
@@ -19,10 +19,6 @@ module Restforce
       def cache
         options[:cache]
       end
-
-      def params_to_ignore
-        (super + ["Auth"]).uniq
-      end
     end
   end
 end
--- a/lib/restforce/middleware/caching.rb
+++ b/lib/restforce/middleware/caching.rb
@@ -12,13 +12,17 @@ module Restforce
     # We don't want to cache requests for different clients, so append the
     # oauth token to the cache key.
     def cache_key(env)
-      super(env) +
-        env[:request_headers][Restforce::Middleware::Authorization::AUTH_HEADER].
-          gsub(/\s/, '')
+      super(env) + hashed_auth_header(env)
     end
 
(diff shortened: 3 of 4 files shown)


timrogers:
Good spot :+1: Can you get the tests passing, then I'm happy to merge this?

atmos:
It looks like we're relying on a gem, public_suffix, that won't install on 1.9.x. Can we remove those entries from travis.yml or is there a more restrictive version we can pin to?

timrogers:
@atmos From looking at the gems history, it looks like versions before 2.0.4 should work - could you have a play?

atmos:
@timrogers I opened up #274 to show off the build failure as not being relevant to this PR.

timrogers:
Not sure how best to handle this dependency problem. What do you think @ccutrer? You've been helpful on this kind of thing before!

web-connect:
My meager vote goes to deprecating 1.9

ccutrer:
The "proper" way is for you to have a more lenient version requirement in the restforce gemspec, and then anyone using Ruby 1.9, in their Gemfile needs to require a version of public_suffix old enough that it will work on 1.9 (they can't rely on bundler not ruling out newer versions that don't work on older ruby).

atmos:
Yeah, this is obviously not ideal since it's a security update and it'd be nice if people could get this even if they're on older versions of this gem that worked with 1.9.x.

AFAICT `public_suffix` started enforcing `>= 1.9.3` with version [1.4.3](https://rubygems.org/gems/public_suffix/versions/1.4.3) on June 12th, 2014. It started enforcing `>= 2.0` in version [1.5.0](https://rubygems.org/gems/public_suffix/versions/1.5.0) on March 24th, 2015. So this doesn't look like it has anything to do with a newer version of public_suffix since its been broken for quite some time.

The `Gemfile.lock` file that generates from this PR provides this:

```yaml
GEM
  remote: https://rubygems.org/
  specs:
    addressable (2.5.0)
      public_suffix (~> 2.0, >= 2.0.2)
```

So maybe it's the newer versions of addressable? It looks like [2.5.0](https://rubygems.org/gems/addressable/versions/2.5.0) came out on November 4th, 2016 and introduced this dependency.

This leaves us with the wonderful situation of continuing to support 1.9.3 or not follow the latest stable release of addressable. Since [ruby 1.9.3 isn't maintained anymore](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/) my vote is for dropping support for it and bumping the minor version here.

timrogers:
Matches up with my thoughts - I'd be pro moving to Ruby 2+ compatibility.

I'll look to do this in a few days but will leave some time for comments -
I invite anyone to leave a message here or drop me an email if you very
strongly disagree!

atmos:
I updated #274 to document the deprecation of 1.9.x.

timrogers:
@atmos I'll get on this once I've had a chance to go through the Ruby 2+ PR - sorry for the delay!

timrogers:
@atmos Can you rebase your branch from `master`? (I'd do it myself but it's pretty messy to do that way!). We'll the release this as 2.5.0.

atmos:
@timrogers All set.

PR #274, opened 2016-11-29
Deprecate ruby 1.9.x support

atmos:
This is a PR with no changes to show off that master doesn't build due to something changing in the gem dependencies. 

/cc https://github.com/ejholmes/restforce/pull/273

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,9 @@
 language: ruby
 rvm:
-  - 1.9.3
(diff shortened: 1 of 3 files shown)


atmos:
As [discussed in 273](https://github.com/ejholmes/restforce/pull/273#issuecomment-264017136) this deprecates support for ruby 1.9.x which was EOL'd in February 2015.

PR #275, opened 2016-11-30
Add a Tooling API example to the README

davebrace:
As per requested in https://github.com/ejholmes/restforce/issues/268, add an example that shows how the Tooling API can be used.

The example creates a "Twitter Username" field on the standard "Account" object.


timrogers:
:ok_hand:

Issue #263, opened 2016-10-19
Upsert failing due to PATCH not being an acceptable method


cworsley4:
I'm calling upsert on a custom object, where I'm updating based on Id. In the case in question, upsert should perform an insertion since Id is nil. That is the behavior I'm expecting. The SFDC error message reads: `METHOD_NOT_ALLOWED: HTTP Method 'PATCH' not allowed. Allowed are GET,HEAD,POST`

timrogers:
@cworsley4 Can you give an example of your code? I'm wondering if it might be an example of "Inserting New Records Using Id as the External ID" at https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm, which seems to use `POST`.

PR #277, opened 2016-12-05
Add support for upsert with Id field

ecbypi:
Simplify extracting `external_id` from `attrs` hash by using
`Hash#delete` with all possible transformations of `field`. Avoids
iterating through `attrs` twice.

Addresses #263.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -361,12 +361,17 @@ module Restforce
       # error returned if the external ID provided matches multiple records (in which
       # case the conflicting IDs can be found by looking at the response on the error)
       def upsert!(sobject, field, attrs)
-        external_id = attrs.
-          fetch(attrs.keys.find { |k, v| k.to_s.downcase == field.to_s.downcase }, nil)
-        attrs_without_field = attrs.
-          reject { |k, v| k.to_s.downcase == field.to_s.downcase }
-        response = api_patch "sobjects/#{sobject}/#{field}/#{URI.encode(external_id)}",
-                             attrs_without_field
+        external_id =
+          extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
+
+        response =
+          if field.to_s == "Id" && (external_id.nil? || external_id.strip.empty?)
+            version_guard(37.0) do
+              api_post "sobjects/#{sobject}/#{field}", attrs
+            end
+          else
+            api_patch "sobjects/#{sobject}/#{field}/#{URI.encode(external_id)}", attrs
+          end
 
         (response.body && response.body['id']) ? response.body['id'] : true
       end
@@ -482,6 +487,14 @@ module Restforce
                                  "see https://github.com/ejholmes/restforce/blob/master" \
                                  "/README.md#api-versions"
         end
+      end
+
+      def extract_case_insensitive_string_or_symbol_key_from_hash!(hash, key)
+        value = hash.delete(key.to_sym)
+        value ||= hash.delete(key.to_s)
+        value ||= hash.delete(key.to_s.downcase)
+        value ||= hash.delete(key.to_s.downcase.to_sym)
+        value
       end
 
       # Internal: Errors that should be rescued from in non-bang methods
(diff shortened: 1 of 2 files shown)


lib/restforce/concerns/api.rb
```diff
@@ -361,12 +361,17 @@ def upsert(*args)
       # error returned if the external ID provided matches multiple records (in which
       # case the conflicting IDs can be found by looking at the response on the error)
       def upsert!(sobject, field, attrs)
-        external_id = attrs.
-          fetch(attrs.keys.find { |k, v| k.to_s.downcase == field.to_s.downcase }, nil)
-        attrs_without_field = attrs.
-          reject { |k, v| k.to_s.downcase == field.to_s.downcase }
-        response = api_patch "sobjects/#{sobject}/#{field}/#{URI.encode(external_id)}",
-                             attrs_without_field
+        external_id = attrs.delete(field.to_sym)
```
timrogers: It might be nice to extract this into a private method rather than have this rather verbose piece of code here. I guess it'd be something like `extract_case_insensitive_string_or_symbol_key_from_hash`. What do you think?

timrogers: Could we otherwise not document that you need to provide symbol keys and a symbol field name? That feels a bit nicer than having to do these comparisons every time.

ecbypi: > It might be nice to extract this into a private method rather than have this rather verbose piece of code here. I guess it'd be something like `extract_case_insensitive_string_or_symbol_key_from_hash`. What do you think?

If it won't be used elsewhere, my inclination is to leave it in `upsert!` to avoid indirection when debugging and add comment to describe what's going on.  I'm open to whichever you think makes sense long term.  

> Could we otherwise not document that you need to provide symbol keys and a symbol field name? That feels a bit nicer than having to do these comparisons every time.

The flexibility is nice for consumers of the gem to accommodate whatever sources `field` and `attrs` are coming from and these changes reduce the number of comparisons performed from `attrs.length` to a max of 4 if the match is found on the last check.  Changing the current behavior might be a pain for anyone incidentally relying on it.

timrogers:
I've just noticed that `upsert` with `Id` as the external ID is only available in API versions 37 and later (see the docs [here](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm)). By default, we use a much older version for backwards compatibility.

We could wrap this in a version guard, but that'll be very ugly given that we've got two code paths here. For me, that pushes me towards wanting this to be a separate method. What do you think?

ecbypi:
> We could wrap this in a version guard, but that'll be very ugly given that we've got two code paths here. For me, that pushes me towards wanting this to be a separate method. What do you think?

The version guard would only go in the code path that uses `Id` for the attribute so it doesn't add any new code paths within the method. I think it's fine if the implementation is a slightly more complicated if it has sufficient documentation and provides better usability for gem consumers (having one method to call for upsert functionality).

timrogers:
@ecbypi Sorry for the delay in getting back to you! Jumping on this now.

lib/restforce/concerns/api.rb
timrogers: I think I'd rather go for the `#extract_case_insensitive_string_or_symbol_key_from_hash` route - it adds a little indirection, but I'd always (personally) tend to want to add a method with a really clear, expressive name rather than a comment.

It might not be reused for now, but when I read the code, I'll instantly know what we're trying to do.

Otherwise, happy with supporting both strings and symbols :+1:

timrogers:
Let's add the version guard described above too, and then I'm good with this! Thanks for your patience 🌟

ecbypi:
@timrogers added the API version guard and extracted the private method.

timrogers:
@ecbypi Got a CI failure here due to this not passing Rubocop's rules - can you take a look? If you don't have a chance in the next few days, I'll fix and merge myself :smile: (or feel free to comment back if you'd rather I do that!)

ecbypi:
@timrogers will take a look tonight.

ecbypi:
@timrogers apologies for the delay, I've addressed the rubocop violations.

PR #278, opened 2016-12-14
Allow upgrading to faraday 0.10

alexluke:


restforce.gemspec
```diff
@@ -18,8 +18,8 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.0'
 
-  gem.add_dependency 'faraday', '~> 0.9.0'
-  gem.add_dependency 'faraday_middleware', '>= 0.8.8'
+  gem.add_dependency 'faraday', ['>= 0.9.0', '< 1.0']
```
timrogers: Not sure what the best way to bind the versions is - given that they've gone for a `0.10.0`, I suspect there'll be a `0.11.0` after that. `1.0` could still be a long way away!

Could we perhaps bind like this - that way, we can guarantee getting versions that we know will work:

```ruby
gem.add_dependency 'faraday', ['>= 0.9.0', '<= 0.10.0']
```

restforce.gemspec
```diff
@@ -18,8 +18,8 @@ Gem::Specification.new do |gem|
...
+  gem.add_dependency 'faraday_middleware', ['>= 0.8.8', '< 1.0']
```
timrogers: As above 👍

alexluke:
I went up to 1.0 to match how `faraday_middleware` did it, but I agree that it's better to be on the safe side.

timrogers:
For whatever reason, Travis hasn't run this, but I've tested it locally and it works. Merging now!

@alexluke Thanks :heart:

PR #281, opened 2016-12-29
Custom Header middleware

moskeyombus:
## Problem
The Restforce gem does not currently support custom HTTP request headers out of the box.  It's currently possible for users to define their own middleware in their application to handle request header injection (discussion here: https://github.com/ejholmes/restforce/issues/109).  While this works, there is logic in Salesforce API interactions that can be triggered based on the presence of custom headers, so it is justifiable to support this in Restforce.

## Solution
Added `Restforce::Middleware::CustomHeaders` to the gem.  This allows user to pass a hash of custom headers values into the client when initializing.  If a valid hash is provided, the values in the hash are injected into the headers of every HTTP request made by that specific client instance.

Here's an example (I added this to the README as well):
```ruby
client = Restforce.new(oauth_token: 'access_token',
                       instance_url: 'instance url',
                       api_version: '38.0',
                       request_headers: { 'sforce-auto-assign' => 'FALSE' })

```

## Details
A couple things:
- It is technically possible to specify custom headers with same keys as pre-existing Authorization headers.  The custom headers will overwrite the pre-existing Auth headers, I'm ok with changing the code to prefer the Auth headers and never overwrite them, although the argument could be made that users may _want_ to override those headers for whatever reason.
- I added documentation for this argument but I'm no copywriter, feel free to suggest changes there.

### DIFF
--- a/lib/restforce/middleware/custom_headers.rb
+++ b/lib/restforce/middleware/custom_headers.rb
@@ -1,0 +1,11 @@
+module Restforce
+  # Middleware that allows you to specify custom request headers when initializing Restforce client
+  class Middleware::CustomHeaders < Restforce::Middleware
+    def call(env)
+      headers = @options[:request_headers]
+      env[:request_headers].merge!(headers) if headers.is_a?(Hash)
+
+      @app.call(env)
+    end
+  end
+end
--- a/lib/restforce/middleware/custom_headers.rb
+++ b/lib/restforce/middleware/custom_headers.rb
@@ -1,5 +1,6 @@
 module Restforce
-  # Middleware that allows you to specify custom request headers when initializing Restforce client
+  # Middleware that allows you to specify custom request headers
+  # when initializing Restforce client
   class Middleware::CustomHeaders < Restforce::Middleware
     def call(env)
       headers = @options[:request_headers]
--- a/README.md
+++ b/README.md
@@ -194,6 +194,23 @@ to do some custom error handling. The bang methods will raise exceptions, while 
 non-bang methods will return false in the event that an exception is raised. This
 works similarly to ActiveRecord.
 
+
+### Custom Headers
+
+Salesforce allows the addition of
+[custom headers](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers.htm)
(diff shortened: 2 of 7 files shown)


panozzaj:
Looks pretty solid to me! +1

lib/restforce/middleware/custom_headers.rb
```diff
@@ -0,0 +1,12 @@
+module Restforce
+  # Middleware that allows you to specify custom request headers
+  # when initializing Restforce client
+  class Middleware::CustomHeaders < Restforce::Middleware
+    def call(env)
+      headers = @options[:request_headers]
+      env[:request_headers].merge!(headers) if headers.is_a?(Hash)
```
timrogers: Would it make sense to merge the specified `request_headers` on top of `headers`? That would allow you to override which feels like the right behaviour.

moskeyombus: As it stands right now, we are allowing the user to overwrite whatever existing headers are present in the request (`env[:request_headers]`) with anything they pass in as `@option[:request_headers]`.  Are you saying we should flip it to 
```
headers.merge!(env[:request_headers])
```
so the user can't overwrite headers set elsewhere in the gem?

I'm totally fine with this, honestly I could argue in favor of going either direction, just making sure I understand what you're suggesting

timrogers:
Ah,  just misread the code (even after checking several times!) - it already does exactly what I'm proposing. I'll merge this later today.

timrogers:
This will be released ASAP.

jonnymacs:
@timrogers @panozzaj @moskeyombus the logger middleware should be included last in the Faraday connection builder (after gzip and custom headers) else neither gzip nor any custom headers get logged. This leads to confusion. I can make a PR...

timrogers:
@jonnymacs Good spot! Fancy making a PR?

timrogers:
Might be nice to add some kind of test as well that ensures that the logger middleware comes last, if we can.

jonnymacs:
@timrogers sure thing. i'll tackle the test coverage as well

timrogers:
Thanks :heart:

moskeyombus:
@jonnymacs good catch, thanks for taking care of this!

PR #293, opened 2017-03-17
enhancement to #281. ensure logger middleware is used last

jonnymacs:
to ensure all headers are logged.

also added request_headers global config support

### DIFF
--- a/lib/restforce/config.rb
+++ b/lib/restforce/config.rb
@@ -141,6 +141,9 @@ module Restforce
     # Set SSL options
     option :ssl, default: {}
 
+    # A Hash that is converted to HTTP headers
+    option :request_headers
+
     # Set a logger for when Restforce.log is set to true, defaulting to STDOUT
     option :logger, default: ::Logger.new(STDOUT)
 
--- a/spec/unit/concerns/connection_spec.rb
+++ b/spec/unit/concerns/connection_spec.rb
@@ -60,6 +60,22 @@ describe Restforce::Concerns::Connection do
         end
       end
     end
+
+    describe ":logger option" do
+      let(:options) { { adapter: Faraday.default_adapter } }
+
+      before(:each) do
+        client.stub(:authentication_middleware)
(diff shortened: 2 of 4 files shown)


lib/restforce/config.rb
```diff
@@ -141,6 +141,9 @@ def option(*args)
     # Set SSL options
     option :ssl, default: {}
 
+    # A Hash that is converted to http headers
```
timrogers: Can you tweak this to HTTP?

spec/unit/concerns/connection_spec.rb
```diff
@@ -60,6 +60,22 @@
         end
       end
     end
+
+    describe ":logger" do
```
timrogers: Could we change this to `":logger option"` to match the above `describe` block?

spec/unit/concerns/connection_spec.rb
```diff
@@ -60,6 +60,22 @@
...
+
+      before(:each) do
+        client.stub(:authentication_middleware)
+        client.stub(:cache)
+        client.stub(options: options)
+        Restforce.stub(log?: true)
+      end
+
+      it "must always be used last before the Faraday Adapter" do
+        client.middleware.handlers.pop(2).index(Restforce::Middleware::Logger).
```
timrogers: It'd probably be nicer here to avoid `Array#pop` and to more directly check that it's the penultimate one? (e.g. `client.middleware.handlers.reverse.index(Restforce::Middleware::Logger).should eq(1)`)

jonnymacs: no like Array#pop? :) no prob. reverse is a nice trick here that didn't come to me.

jonnymacs:
@timrogers ready to merge?

timrogers:
@jonnymacs Live in v2.5.2!

jonnymacs:
sweet! thanks @timrogers

PR #296, opened 2017-04-12
Do not modify attributes hash in upsert! method

velveret:
Duplicate the hash before modification

https://github.com/ejholmes/restforce/issues/159

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -361,6 +361,7 @@ module Restforce
       # error returned if the external ID provided matches multiple records (in which
       # case the conflicting IDs can be found by looking at the response on the error)
       def upsert!(sobject, field, attrs)
+        attrs = attrs.dup
         external_id =
           extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
 
--- a/spec/unit/concerns/api_spec.rb
+++ b/spec/unit/concerns/api_spec.rb
@@ -282,116 +282,123 @@ describe Restforce::Concerns::API do
     end
   end
 
-  describe '.create!' do
-    let(:sobject)    { 'Whizbang' }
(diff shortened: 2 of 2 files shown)


lib/restforce/concerns/api.rb
```diff
@@ -360,9 +360,13 @@ def upsert(*args)
       # Raises an exception if an error is returned from Salesforce, including the 300
       # error returned if the external ID provided matches multiple records (in which
       # case the conflicting IDs can be found by looking at the response on the error)
-      def upsert!(sobject, field, attrs)
+      def upsert!(sobject, field, attrs_arg)
+        attrs = attrs_arg.dup
```
timrogers: Feels a bit ugly having the argument named `attrs_arg` - could we just do the following? Should work fine I think? Just tried and it worked fine for me playing in `irb`.

```ruby
  attrs = attrs.dup
  # ...
end

File: lib/restforce/concerns/api.rb [hunk 2]
```diff
@@ -360,9 +360,13 @@ def upsert(*args)
...
         external_id =
           extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
+        if field.to_s != "Id" && (external_id.nil? || external_id.strip.empty?)
```
timrogers: I think I'd prefer to leave this out for now, or at least to do it in a separate PR to keep things clear and simple.

timrogers:
💖

velveret:
Amended the commit based on the suggestions. Thanks!

timrogers:
Looks good :+1: Specs are currently failing in Travis for some reason... I'll merge this and look at that another time.

lib/restforce/concerns/api.rb
velveret: Opened follow-up PR in https://github.com/ejholmes/restforce/pull/297

PR #297, opened 2017-04-19
Raise error in upsert! when external id is missing

velveret:
Discovered as a result of https://github.com/ejholmes/restforce/issues/159 - when the external id was missing after being deleted, the error surfaced as `NoMethodError·undefined method 'gsub' for nil:NilClass` since `upsert!` would try to URI encode `nil` (https://github.com/ejholmes/restforce/blame/master/lib/restforce/concerns/api.rb#L374)

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -364,6 +364,9 @@ module Restforce
         attrs = attrs.dup
         external_id =
           extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
+        if field.to_s != "Id" && (external_id.nil? || external_id.strip.empty?)
+          raise ArgumentError, 'Specified external ID field missing from provided attributes'
+        end
 
         response =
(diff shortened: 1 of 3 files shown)


lib/restforce/concerns/api.rb
```diff
@@ -364,6 +364,9 @@ def upsert!(sobject, field, attrs)
         attrs = attrs.dup
         external_id =
           extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
+        if field.to_s != "Id" && (external_id.nil? || external_id.strip.empty?)
+          raise ArgumentError, 'External id field missing from attrs.'
```
timrogers: Can we change this to "Specified external ID field missing from provided attributes"?

timrogers:
@velveret Thanks! I'll release this with #299 once that's merged.

timrogers:
Really wish I knew why the tests were red 😭

Issue #238, opened 2016-05-24
Salesforce disabling TLS 1.0, is Restforce compatible with TLS 1.1 or 1.2?


samlehman:
https://help.salesforce.com/apex/HTViewSolution?urlname=Salesforce-disabling-TLS-1-0&language=en_US&

antwonlee:
Does TLS have to do more with updating the browser and your https SSL protocol support?  The Heroku SSL at least says it supports TLS 1.0, 1.1, 1.2  https://devcenter.heroku.com/articles/ssl-beta#ssl-details

theSteveMitchell:
Restforce relies on Faraday, which supports multiple HTTP Adapters.  It will use Net::HTTP by default.  

According to this SO answer, TLS 1.2/1.1 is supported in Nett::Http in Ruby 2 +, not in 1.9.3. http://stackoverflow.com/questions/11059059/is-it-possible-to-enable-tls-v1-2-in-ruby-if-so-how

If you're using another adapter, like Typhoeus, SSL is handled by Libcurl (should be fine).  

I recommend you test with https://help.salesforce.com/apex/HTViewSolution?urlname=Salesforce-disabling-TLS-1-0&language=en_US&#InboundIntegrationTest

bvlaar:
This weekend some Sandbox orgs were updated to[ require TLS 1.1+](https://help.salesforce.com/apex/HTViewSolution?id=000221207#Whatischange).

I'm using `Restforce v2.2.1` and `Ruby 2.0.0`. All of my requests are returning the following message this morning:

```Faraday::ClientError (UNSUPPORTED_CLIENT: TLS 1.0 has been disabled in this organization. Please use TLS 1.1 or higher when connecting to Salesforce using https.)```

Has anyone solved this issue and figured out how to require TLS 1.1 when making these requests?

bvlaar:
For anyone else that experiences the same issue, any version of Ruby > 2.0.0 includes TLSv1.1+. My fix was as simple as upgrading my Ruby version.

veelenga:
Successfully made it to work with [Typhoeus](https://github.com/typhoeus/typhoeus) and ruby 1.8.7. Short sample:

```ruby
auth = Typhoeus.post "#{url}/services/oauth2/token",
  :headers => { 'Accept' => 'application/json' },
  :body => {
    :client_id     => "#{client_id}",
    :grant_type    => "#{password}",
    :client_secret => "#{client_secret}",
    :username      => "#{username}",
[...]

PR #299, opened 2017-04-24
Use SSL options for authentication request

jvdp:
This change ensures that the SSL options passed to the request are also used for the authentication request (if it is needed) in addition to the actual resource request. This allows enforcement of an SSL version (eg. TLS v1.2) for increased security.

See also: #238

### DIFF
--- a/lib/restforce/middleware/authentication.rb
+++ b/lib/restforce/middleware/authentication.rb
@@ -81,7 +81,8 @@ module Restforce
 
     def faraday_options
       { url: "https://#{@options[:host]}",
-        proxy: @options[:proxy_uri] }.reject { |k, v| v.nil? }
+        proxy: @options[:proxy_uri],
+        ssl: @options[:ssl] }
     end
   end
 end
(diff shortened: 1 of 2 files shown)


timrogers:
Looks super reasonable to me on the surface. I'll review properly tomorrow
- thanks!

lib/restforce/middleware/authentication.rb
```diff
@@ -81,7 +81,8 @@ def encode_www_form(params)
 
     def faraday_options
       { url: "https://#{@options[:host]}",
-        proxy: @options[:proxy_uri] }.reject { |k, v| v.nil? }
+        proxy: @options[:proxy_uri],
+        ssl: @options[:ssl] }.reject { |k, v| v.nil? }
```
timrogers: Do we need this `Hash#reject`? We don't have it in `Restforce::Concerns::Connection#connection_options`, and it seems to work fine.

jvdp: Sure, why not.

jvdp:
Thanks for the fast merge and release! 👍

jcabasc:
Hey @jvdp I'm curious what's the proper way to send the version, as a symbol `:TLSv1_2` as a string `'TLSv1.1'` or it does not matter?. Thanks.

jvdp:
@jcabasc I'm using `config.ssl = { version: :TLSv1_2 }`, not sure if it matters.

PR #300, opened 2017-04-25
Fix Faraday version for development and testing to v0.11.0

timrogers:
The gem is compatible with versions of Faraday up to v1.0 (yet to be released! - currently the latest is v0.12.1) but at present, our tests only work with v0.11.0 and earlier (the current version is v0.12.1, so we're not too far behind).

This updates our `Gemfile`s to lock to this version so tests will pass again in Travis. I'll look as soon as I can at fixing the tests for `Restforce::Concerns::Connection` so they can work with newer Faraday versions - looks like its API has changed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ gemspec
 
 gem 'rake'
 gem 'jruby-openssl', :platforms => :jruby
+gem 'faraday', '~> 0.11.0'
 
 group :development do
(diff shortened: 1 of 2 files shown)


timrogers:
A fairly uncontroversial change which fixes our tests until we work out a longer term solution - just making this into a PR by way of a record.

Issue #301, opened 2017-05-02
Documentation for retrieving a Document is in error


tfuda:
I tried following the example here: https://github.com/ejholmes/restforce#documents to download the body of a Document and save it to a file. This didn't work for me. Examination of document.Body shows that it contains a URL, like this: /services/data/v35.0/sobjects/Document/0155C00000053lWQAQ/Body for retrieving the body. To correctly retrieve the document body you must do the following:

```
document = client.query("SELECT Id, Name, Body FROM Document WHERE FolderId = '#{folder.Id}').first
# Get the resource at the URL returned as "Body". This returns a Faraday::Response object
resp = ptech_client.get(document.Body)
File.open(document.Name, 'wb') { |f| f.write(resp.body) }

timrogers:
This doesn't seem to match up with what is mentioned in the Salesforce docs - see https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_document.htm.

daveespo:
This appears to be a bug (or, I'm sure they'll call it 'working as designed') in the Salesforce REST API

If you run a SOQL query through the API like:

````
SELECT Id,Name,Body FROM Document LIMIT 2002
````

(the LIMIT 2002 ensures that you get a nextRecordsUrl)

The first 'page' worth of records just has the URL to the Document Body field
 
![2017-05-03_1301](https://cloud.githubusercontent.com/assets/233309/25672249/c0c50326-3000-11e7-8c16-629f8c83cd16.png)


But when you retrieve the 'nextRecordsUrl' for that same query, you see that only a single record is returned and it contains the actual binary content of the Body field

![2017-05-03_1255](https://cloud.githubusercontent.com/assets/233309/25671994/f3092728-2fff-11e7-9760-26c1e8d1e0bc.png)

And notice that the 'nextRecordsUrl' is for offset 2001, not 4000 as you'd expect for a normal query result

This behavior of only returning the Body content if a single record is retrieved is hinted at here:

https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_attachment.htm

> You can access all of the fields using a query() call. However, you can't receive the Body field for multiple records in a single query() call. If your query returns the Body field, your client application must ensure that only one row with one Attachment is returned; otherwise, an error occurs. A more effective approach is to return IDs (but not Attachment records in the Body field) from a query() call and then pass them into retrieve() calls that return the Body field.

[...]

PR #302, opened 2017-05-03
#301 - Documents nuanced behavior in REST API for BLOB field retrieval

daveespo:
Hopefully this covers the workaround for #301


timrogers:
Thanks! ❤️

PR #304, opened 2017-05-05
Add #empty? to Collection

derekgottlieb:


lib/restforce/collection.rb
```diff
@@ -31,6 +31,11 @@ def size
     end
     alias_method :length, :size
 
+    # Return whether the collection contains any objects.
```
timrogers: I think this will respond to `#any?` already and do the same thing, since it uses `Enumerable`. Can you check? Nice idea though.

derekgottlieb: Good point. I'd missed that `#empty?` was specifically an `Array` method and not provided by `Enumerable`. I've quickly verified you can call `#any?` without passing a block and it should give you the inverse of `#empty?`. At least, assuming there's no scenario where we could get a collection that only contains `false` or `nil` values, which I can't recall ever running into.

I frequently find myself writing code to handle an empty query result, so I would find the syntax a little more expressive than doing `! #.any?` everywhere. But I can understand if you'd prefer to keep the interface minimal and stick to what `Enumerable` provides unless you've got a compelling reason to define additional methods (like for `#each`).

timrogers: I'd be quite keen to stick with what we have now rather than adding an alias - potentially something you could add as a monkey patch if you like? Or would `#none?` serve your needs better? (At least avoids a `!`...)

derekgottlieb: I'd prefer to avoid monkey patches, but I'm fine with using `#none?`. There is a performance difference between that and the less flexible `#empty?`, but probably not enough to care about if you'd prefer to maintain a minimal interface. 

Quick benchmark comparing a 1000 calls to `#empty?` vs `#none?` for an empty collection and a collection with about a 1000 sobjects:

```
                        user     system      total        real
empty? with empty:  0.000000   0.000000   0.000000 (  0.000584)
none? with empty:   0.000000   0.000000   0.000000 (  0.001356)
empty? with 1000:   0.000000   0.000000   0.000000 (  0.000399)
none? with 1000:    0.050000   0.000000   0.050000 (  0.052230)
```

timrogers: I think I'd rather stick with the minimal interface. Thanks so much for your PR :heart: Please don't be disheartened by me not taking this one - I'm just trying to be opinionated and keep things minimal where we can.

timrogers:
Not merging per [my comment](https://github.com/ejholmes/restforce/pull/304#discussion_r115234695)

lib/restforce/collection.rb
derekgottlieb: 👍

PR #313, opened 2017-07-18
When configuring an API version, check that it's valid

timrogers:
Salesforce API versions are string representations of a number to one decimal place, for example `"38.0"` or `"40.0"`.

As noted by @wmadden [in his comment](https://github.com/ejholmes/restforce/issues/160#issuecomment-256888846), if you set an invalid API version, the Salesforce API just returns an unhelpful 404, and we raise a `Faraday::ResourceNotFound`. It is not terribly clear when this happens the problem is your API version, which gets put in the URL.

This tweaks `Restforce::Configuration::Option` to allow an option to have a validator, which is used to check values provided when they are set. You specify a lambda which is called, and may
raise an error, much like we use for specifying defaults.

We then add a validator for the `api_version` option, comparing the provided value to a regular expression, and raising an `InvalidOptionError` with a helpful message if it's invalid.

### DIFF
--- a/lib/restforce.rb
+++ b/lib/restforce.rb
@@ -42,6 +42,7 @@ module Restforce
   AuthenticationError = Class.new(Error)
   UnauthorizedError   = Class.new(Error)
   APIVersionError     = Class.new(Error)
+  InvalidOptionError  = Class.new(Error)
 
   class << self
     # Alias for Restforce::Data::Client.new
--- a/lib/restforce/config.rb
+++ b/lib/restforce/config.rb
@@ -49,24 +49,27 @@ module Restforce
         @name = name
         @options = options
         @default = options.fetch(:default, nil)
+        @validator = options.fetch(:validator, nil)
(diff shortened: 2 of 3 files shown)


timrogers:
@erlandsona Do you want to give this a first review? 💟

erlandsona:
Thanks a ton for this. I know it would've saved me at least a few hours.

timrogers:
I'm going to merge this as I can't see any potential for issues with it, given that the specs pass and I can configure a client correctly. Before I release this (which I'll aim to do in the next week or so), I'll have a think about other settings we might be able to validate and might tag you in the PRs if I add any.

timrogers:
I'm going to hold off merging for a little bit actually - I just tried instantiating the client with an invalid API version specified and it didn't blow up 🤔 I'll investigate tomorrow!

timrogers:
I'll try to get back on this in the next few days.

PR #315, opened 2017-07-24
Add gemspec metadata

greysteil:
Makes it easy to programatically find the source code and changelog for `restforce`, using the rubygems API. More details on the direction of travel for gemspecs is [here](https://github.com/rubygems/rubygems.org/issues/1127) and [here](https://github.com/rubygems/rubygems.org/pull/1234), and the current state is [here](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification.rb).


greysteil:
@timrogers - quick one for you.

PR #319, opened 2017-08-01
Support for data replication end points 

jonydev:
We've needed to extend to provide support for the data replication end
points. Doing so means upgrading the api version to 29+ and implementing
two end points getUpdatedBetween and getDeletedBetween which take two
Time object representing the beginning and end of a time window you wish
salesforce to provide you notification of updated

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -453,6 +453,16 @@ module Restforce
         api_get(path).body
       end
 
+      def get_updated_between(sobject, startDate, endDate)
+        path = "sobjects/#{sobject}/updated/?start=#{startDate.utc.iso8601}&end=#{endDate.utc.iso8601}"
+        api_get(path).body
+      end
+
+      def get_deleted_between(sobject, startDate, endDate)
+        path = "sobjects/#{sobject}/deleted/?start=#{startDate.utc.iso8601}&end=#{endDate.utc.iso8601}"
+        api_get(path).body
+      end
+
       # Public: Finds recently viewed items for the logged-in user.
       #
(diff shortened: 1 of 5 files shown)


lib/restforce/concerns/api.rb
```diff
@@ -453,6 +453,16 @@ def select(sobject, id, select, field = nil)
         api_get(path).body
       end
 
+      def get_updated_between(sobject, startDate, endDate)
```
timrogers: Can we call these `start_date` and `end_date` to be more Ruby-like?

lib/restforce/config.rb
```diff
@@ -88,7 +88,7 @@ def option(*args)
       end
     end
 
-    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '26.0' }
```
timrogers: I'd prefer not to change the default for compatibility reasons, and instead just use a `version_guard` in your methods - see the `#limits` method in `api.rb` for an example.

spec/unit/concerns/api_spec.rb
```diff
@@ -427,6 +427,36 @@
     end
   end
 
+  describe '.get_updated_between' do
+    let(:sobject) {'Account'}
```
timrogers: Can we add spaces inside the curly braces? (e.g. `{ 'Account' }`) Here and elsewhere!

spec/unit/concerns/api_spec.rb
```diff
@@ -427,6 +427,36 @@
...
+    let(:endTime) {Time.now}
```
timrogers: As above, can we use names like `end_time`?

PR #322, opened 2017-08-07
Undefined method strip when external_id is a Fixnum

AlexandruCD:
If external Id is a Fixnum extract_case_insensitive_string_or_symbol_key_from_hash will return it as a Fixnum too, and we'll get Undefined method strip for Fixnum. Happens for symbols too.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -363,7 +363,7 @@ module Restforce
       def upsert!(sobject, field, attrs)
         attrs = attrs.dup
         external_id =
-          extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field)
+          extract_case_insensitive_string_or_symbol_key_from_hash!(attrs, field).to_s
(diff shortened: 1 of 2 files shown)


v-yarotsky:
this fix is worth backporting, using Fixnum as an external id has been broken since https://github.com/ejholmes/restforce/commit/f61689926418cab4622578a0e8b016fffc606947

```
irb(main):001:0> URI.encode(1)
NoMethodError: undefined method `gsub' for 1:Fixnum
        from /Users/vlad/.rbenv/versions/2.3.3/lib/ruby/2.3.0/uri/rfc2396_parser.rb:305:in `escape'
        from /Users/vlad/.rbenv/versions/2.3.3/lib/ruby/2.3.0/uri/common.rb:100:in `escape'
        from (irb):1
        from /Users/vlad/.rbenv/versions/2.3.3/bin/irb:11:in `<main>'
irb(main):002:0>
```

timrogers:
@AlexandruCD Thanks for taking the time to fix this :heart:

PR #332, opened 2017-08-23
Improved error handler

dmitriy-strukov:
Improved error message, instead false you can see detail message

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -262,8 +262,8 @@ module Restforce
       # Returns false if something bad happens.
       def create(*args)
         create!(*args)
-      rescue *exceptions
-        false
+      rescue Faraday::ClientError => error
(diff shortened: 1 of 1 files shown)


dmitriy-strukov:
@timrogers please review

timrogers:
The purpose of the `#create` method is *not* to raise exceptions - it's for if you want to perform the operation but don't want your code to explode if it goes wrong. You can see that since the code itself calls `#create!` and then rescues.

If you want an exception as your form of control flow (e.g. you want to be able to see exactly how the API failed), you should use the "bang" form of the method, e.g. `#create!` - that will raise an appropriate exception which will give you what you're looking for.

Hope that helps - let me know if you have any further questions :smile: And please do come back and keep offering contributions for Restforce - it's really appreciated.

dmitriy-strukov:
@timrogers Thank you for explanations

PR #348, opened 2017-11-29
Ensure write access by configuration

StefanGreffenius:
Hey, 

a little change request. Since it seems not possible to configure connected apps in Salesforce as read only, it would be awesome to configurate Restforce with a `read_only` flag.

Useful for envirouments where you want to read data without calling `save`, `save!`, `destroy` or `destroy!` on them.

### DIFF
--- a/lib/restforce/sobject.rb
+++ b/lib/restforce/sobject.rb
@@ -22,11 +22,13 @@ module Restforce
     #   account.Name = 'Foobar'
     #   account.save
     def save
+      ensure_write_access
       ensure_id
       @client.update(sobject_type, attrs)
     end
 
     def save!
+      ensure_write_access
       ensure_id
       @client.update!(sobject_type, attrs)
     end
@@ -38,11 +40,13 @@ module Restforce
     #   account = client.query('select Id, Name from Account').first
     #   account.destroy
     def destroy
+      ensure_write_access
       ensure_id
       @client.destroy(sobject_type, self.Id)
     end
 
     def destroy!
+      ensure_write_access
       ensure_id
       @client.destroy!(sobject_type, self.Id)
     end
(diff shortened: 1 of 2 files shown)


lib/restforce/sobject.rb
```diff
@@ -22,11 +22,13 @@ def describe_layouts(layout_id = nil)
     #   account.Name = 'Foobar'
     #   account.save
     def save
+      ensure_write_access
```
timrogers: Would it make sense to hook this into all non-`GET` requests, rather than do it on a method-by-method basis?

lib/restforce/sobject.rb
```diff
@@ -63,5 +67,10 @@ def ensure_id
       return true if self.Id?
       raise ArgumentError, 'You need to query the Id for the record first.'
     end
+
+    def ensure_write_access
+      return true unless Restforce.configuration.read_only
+      raise Restforce::UnauthorizedError, 'You are in read only mode.'
```
timrogers: Might be nice to make a new error class for this to avoid overloaded `UnauthorizedError`.

lib/restforce/sobject.rb
StefanGreffenius: Not sure, `*/services/oauth2/token` seems to be a post.

timrogers: I reckon it might be worth just having an override for those cases - non-`GET` is generally correct.

timrogers:
@StefanGreffenius Any chance you might have some time to come back to this?

StefanGreffenius:
@timrogers Hey, can you give me a hint on how to implement? I tried to hook into the api_methods from `define_verbs :get, :post, :put, :delete, :patch, :head` but i didn't managed to get it working.

PR #350, opened 2017-12-14
Update the most recent API version

RKushnir:


timrogers:
@RKushnir Thanks for the PR. There are challenges with updating the API version in that it can introduce backwards incompatibilities - see https://github.com/restforce/restforce/pull/266#issuecomment-257541314. As such, I'd want to include this change in a major version.

timrogers:
Looks like we should be okay in this case. Salesforce publish a changelog with the changes in each API version:

* [v39.0](https://releasenotes.docs.salesforce.com/en-us/spring17/release-notes/rn_api_rest.htm#rn_api_rest)
* [v40.0](https://releasenotes.docs.salesforce.com/en-us/summer17/release-notes/rn_api_rest.htm#rn_api_rest)
* [v41.0](https://releasenotes.docs.salesforce.com/en-us/winter18/release-notes/rn_api_rest.htm#rn_api_rest)

There are very few significant backwards-incompatible changes. I'll probably release this as a major version anyway just to be safe, but it should be an easy upgrade.

RKushnir:
I see.
If compatibility is generally a concern, do you think it's worth to rephrase the readme? ATM it says _We strongly suggest … to use the most recent API version_.

timrogers:
I wrote that because it is best to use the latest version - but upgrading
could be a little tricky in some cases.

I think it’s good practice generally so have a recent version as the
default, but to warn people in the release noted that they should make sure
changes on Salesforce’s side won’t hit them.

PR #366, opened 2018-02-28
removes leading slash from `api_get` calls

scottolsen:
There is already a slash in the `api_path` method below and the duplicate slashes is causing us to get 404 errors when using `get_updated`. 

https://github.com/restforce/restforce/blob/467be4f6db4db10c64b1ec7fa48cb47de46f548a/lib/restforce/concerns/api.rb#L480


timrogers:
@noahc Thanks for the bump - I'll cut a release ASAP.

PR #386, opened 2018-06-25
Add Collection#next_page_url

ryansch:
Extracted from #158.  We use this to asynchronously fan out queries for all of the pages we need to pull from salesforce.  This allows us to do the actual processing of each page in parallel.

This also adds a corresponding method `Collection#query_next_page` which allows the user to use the url from `next_page_url`.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -187,6 +187,25 @@ module Restforce
         mashify? ? response.body : response.body['records']
       end
 
+      # Public: Returns the next batch of a batched query.
+      #
+      # url - The nextRecordsUrl field from the previous batch response.
+      # options - Hash of options to control the query
+      #
+      # Examples
+      #
+      #   # Find the names of all Accounts
+      #   collection = client.query('select Name from Account')
+      #
+      #   new_collection = client.query_next_page(collection.next_page_url)
+      #
+      # Returns a Restforce::Collection if Restforce.configuration.mashify is true.
+      # Returns an Array of Hash for each record in the result if Restforce.configuration.mashify is false.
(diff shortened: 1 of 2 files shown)


ryansch:
Rebased on master.

michaelize:
Can you give an example of how you guys are fetching all the pages and then enqueuing them into sidekiq jobs?

Assuming you'll have to fetch the first page to get the next page url, are you only processing the fetched data in sidekiq?

michaelize:
@asha-cs Saw your comment in #387.

Correct me if I'm wrong, it seems like that the `nextRecordsUrl` returns a QueryCursor link that could expire or getting released if there's too many requests. If there are more than 10 processes requesting different data set from SFDC, then the nextRecordsUrl returned to some of them may become invalid.

https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_apicursors.htm

asha-cs:
> @asha-cs Saw your comment in #387.
> 
> Correct me if I'm wrong, it seems like that the `nextRecordsUrl` returns a QueryCursor link that could expire or getting released if there's too many requests. If there are more than 10 processes requesting different data set from SFDC, then the nextRecordsUrl returned to some of them may become invalid.
> 
> Would like to point this out.
> 
> https://developer.salesforce.com/docs/atlas.en-us.salesforce_app_limits_cheatsheet.meta/salesforce_app_limits_cheatsheet/salesforce_app_limits_platform_apicursors.htm

Hey @michaelize - looks like the limit is enforced per-user, which should be fine for our use case. I'll bring the info back to my team though, thank you for the resource!

PR #388, opened 2018-06-26
Add authentication_wrapper

ryansch:
Adds support for an authentication wrapper that runs around the
authentication action.  The callable should call/yield to the passed
block to run the authentication.

Replaces #162.


ryansch:
Rebased on master.

PR #392, opened 2018-07-30
Use CGI.escape on ID's passed into the api client

jmdx:


reedloden:
This is CVE-2018-3777.

alagos:
No idea why [CVE-2018-3777](https://github.com/advisories/GHSA-534w-937m-v7x3) don't reflect it, but this patch [was also added to 2.5.4](https://github.com/restforce/restforce/commit/f4570f1a0df0462df593a3a91a195c30cb598382), so it's not necessary to update to >= 3.0.0

PR #411, opened 2018-10-16
Properly URL escape ids with spaces in them

pushups:
The client used to call CGI.escape on ids present in URLs.
This can lead to unexpected behavior where a space in an id
is encoded as a +, but is not decoded by Salesforce leading
to duplicate objects during upsert.

This commit replaces the CGI.escape calls with
ERG::Util.url_encode calls which encode space as %20.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'erb'
 require 'uri'
 require 'restforce/concerns/verbs'
 
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -377,7 +377,7 @@ module Restforce
               api_post "sobjects/#{sobject}/#{field}", attrs
             end
           else
-            api_patch "sobjects/#{sobject}/#{field}/#{CGI.escape(external_id)}", attrs
+            api_patch "sobjects/#{sobject}/#{field}/#{ERB::Util.url_encode(external_id)}", attrs
           end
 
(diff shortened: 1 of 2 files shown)


timrogers:
Thanks for your PR Michael! Do you know of any docs that explain how we’re
meant to do encoding with Salesforce? I want to make sure we get this right
this time.

pushups:
@timrogers - Good question. I haven't been able to find the canonical doc that explains the encoding exactly.

The REST API documents make multiple references to URL-encoding, but they never explain exactly how they will decode them.
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_search.htm?search_text=URL-encoded

The other anecdotal evidence is that in Apex they encode URLs with space as %20 -
https://developer.salesforce.com/forums/?id=906F0000000Bb0MIAS
So hopefully their url decoding respects that as well.

I think this is ultimately a problem with `CGI.escape`.
I ran this script to look at the differences:
https://gist.github.com/pushups/496db02ef597f42054dd0cabb9d27e45

It looks like the main difference is `CGI.escape` converts space to +. `ERB::Util::url_encode` looks like it has all the same behavior except converts space to %20.

I manually tested this change against a sandbox salesforce instance and saw that upserting an object with  spaces in the key `'foo bar'` worked as expected and created the proper id in salesforce. (Would previously get `'foo+bar'`)

Let me know if you'd like more information and I can dig around some more, or if there are any other changes you'd like to see to the code or anything else. This library has been super helpful to us, so I'm really glad to have the opportunity to contribute back.

timrogers:
Happy to go with this. I'll get a release done soon.

pushups:
Thanks!

PR #436, opened 2019-01-29
sObject collections support

yukas:


antwonlee:
200 records in one request.  @yukas what made you want to add this?  Any specific scenarios?  Trying to minimize requests?

spectator:
@antwonlee Yes, this is to minimize number of requests that counts against SF request quota. sObject collection request still counted as one request which might significantly reduce your quota usage.

spectator:
@timrogers I was wondering if one of the maintainers can have another look at this pull request? This is part of Salesforce API, it does help a lot with taking number of requests that count against the quota down and it is well tested in production since this pull request was opened (two of our production apps use these methods primarily to sync with Salesforce).

I'm happy to update/reopen it with recent master and make any other changes if any.

michaelize:
Would it be better to create a new file to encapsulate the sobject_collection methods? There's one for the batch API: https://github.com/restforce/restforce/blob/master/lib/restforce/concerns/batch_api.rb

cjlarose:
I also have a use case that would benefit from sobject collection methods.

Specifically, I'm interested in doing [upsert operations with sobject collections][0], but those don't appear to be implemented here.

[0]: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_upsert.htm

igorlvicente:
Can I help moving this forward?

timrogers:
Absolutely! If you want to rebase this and get the tests passing, I'd be happy to test it.

PR #448, opened 2019-04-25
Name exceptions for the error codes that might be returned from Salesforce.com

boblail:
Currently, Restforce raises `Faraday::Error::ClientError` for nearly every kind of error that can occur. Errors like `INVALID_OPERATION` are problems in client code that require human intervention to fix while errors like `REQUEST_LIMIT_EXCEEDED` or `SERVER_UNAVAILABLE` are transient errors that could be handled programmatically by queuing work and retrying later.

This pull request unwraps the various error responses that a client may receive from Salesforce so that API consumers can more easily differentiate errors.

### DIFF
--- a/lib/restforce.rb
+++ b/lib/restforce.rb
@@ -45,6 +45,24 @@ module Restforce
   UnauthorizedError   = Class.new(Error)
   APIVersionError     = Class.new(Error)
 
+  # Inherit from Faraday::Error::ResourceNotFound for backwards-compatibility
+  # Consumers of this library that rescue and handle Faraday::Error::ResourceNotFound
+  # can continue to do so.
+  NotFoundError       = Class.new(Faraday::Error::ResourceNotFound)
+
+  # Inherit from Faraday::Error::ClientError for backwards-compatibility
+  # Consumers of this library that rescue and handle Faraday::Error::ClientError
+  # can continue to do so.
+  ResponseError       = Class.new(Faraday::Error::ClientError)
+  MatchesMultipleError= Class.new(ResponseError)
+  EntityTooLargeError = Class.new(ResponseError)
+
+  module ErrorCode
+    def self.const_missing(constant_name)
+      const_set constant_name, Class.new(ResponseError)
+    end
+  end
+
   class << self
     # Alias for Restforce::Data::Client.new
     #
--- a/lib/restforce/middleware/raise_error.rb
+++ b/lib/restforce/middleware/raise_error.rb
@@ -6,18 +6,22 @@ module Restforce
       @env = env
       case env[:status]
       when 300
-        raise Faraday::Error::ClientError.new("300: The external ID provided matches " \
-                                              "more than one record",
-                                              response_values)
+        raise Restforce::MatchesMultipleError.new(
+          "300: The external ID provided matches more than one record",
+          response_values
+        )
       when 401
         raise Restforce::UnauthorizedError, message
       when 404
-        raise Faraday::Error::ResourceNotFound, message
+        raise Restforce::NotFoundError, message
       when 413
-        raise Faraday::Error::ClientError.new("413: Request Entity Too Large",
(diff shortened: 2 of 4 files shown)


boblail:
I was unable to find a complete list of errorCodes in [the REST API documentation](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/errorcodes.htm) so I allowed exceptions for error codes to be dynamically defined.

Is a complete list of errorCodes documented somewhere? If that list exists and changes relatively infrequently, I'd hardcode the constant names for those exceptions.

boblail:
Hey @timrogers 👋 Sorry to ping you. 😬 Are you still maintaining this?

lib/restforce/middleware/raise_error.rb
```diff
@@ -26,7 +26,10 @@ def on_complete(env)
     end
 
     def message
-      "#{body['errorCode']}: #{body['message']}"
+      message = "#{body['errorCode']}: #{body['message']}"
+      message << "\nRESPONSE: #{JSON.dump(@env[:body])}"
```
timrogers: What would you think about attaching the raw body to the exception as an instance variable with accessor, rather than adding it to the message?

lib/restforce.rb
```diff
@@ -45,6 +45,24 @@ module Tooling
...
+  # Inherit from Faraday::Error::ResourceNotFound for backwards-compatibility
+  # Consumers of this library that rescue and handle Faraday::Error::ResourceNotFound
+  # can continue to do so.
+  NotFoundError       = Class.new(Faraday::Error::ResourceNotFound)
+
+  # Inherit from Faraday::Error::ClientError for backwards-compatibility
+  # Consumers of this library that rescue and handle Faraday::Error::ClientError
+  # can continue to do so.
+  ResponseError       = Class.new(Faraday::Error::ClientError)
+  MatchesMultipleError= Class.new(ResponseError)
```
timrogers: The spacing looks a bit odd here

lib/restforce.rb
```diff
@@ -45,6 +45,24 @@ module Tooling
...
+  EntityTooLargeError = Class.new(ResponseError)
+
+  module ErrorCode
+    def self.const_missing(constant_name)
```
timrogers: I have a natural aversion to `const_missing` - maybe we could use `ResponseError` and just attach an instance variable to it exposing the error code?

lib/restforce/middleware/raise_error.rb
boblail: 👍 Yep! I could do that.

lib/restforce.rb
boblail: Do you have a preference between me indenting everything further by one space or just removing the justify-alignment of `=`s in this diff?

lib/restforce.rb
boblail: What's your aversion to `const_missing`? Is there a danger to it that we could mitigate?

Exposing the error code with an instance variable is fine doesn't really achieve what the constants do. With separate classes for each of Salesforce's error codes:
 1. You can `rescue` from specific error codes without resorting to helper methods or metaprogramming
 2. It's friendlier to Sentry and Airbrake's fingerprinting algorithms. (Sentry uses exception class + backtrace (I believe) to fingerprint errors and, when they are all have the same class name (before this PR), Sentry lumps together exceptions caused by _very different_ Salesforce errors. 😬)

* * *

We could move the guts of `const_missing` to `exception_class_for_error_code`. _That_ method could end like this:

```ruby
if Restforce::ErrorCode.const_defined?(constant_name)
  Restforce::ErrorCode.const_get(constant_name)
else
  Restforce::ErrorCode.const_set(constant_name, Class.new(Restforce::ResponseError))
end
```

Would that be better?

boblail:
👏 Thanks!

Issue #477, opened 2019-09-09
Response not being set correctly on 401 or 404 errors


michaeldbianchi:
When SF is returning a 401 or a 404, the response object is not being set correctly, leading to an inability to programmatically set response status for these status codes. 

Issue https://github.com/restforce/restforce/issues/165 fixed this for all errors besides 401 and 404.

https://github.com/restforce/restforce/blob/9cb413d37bcbdec691d38d4e7468c863c5197067/lib/restforce/middleware/raise_error.rb#L12-L15

I'm willing to submit a PR if there is not a specific reason for omitting the response object.

PR #478, opened 2019-09-09
Bugfix/standardize middleware errors

michaeldbianchi:
Purpose of this PR is to appropriately set the response attribute on the Error raised in middleware. The current state of things are that `401` and `404` status codes fail to set the response and make it impossible to programmatically access the response status, headers, and body of the SF request

Fixes https://github.com/restforce/restforce/issues/477

⚠️ Modifies inheritance of UnauthorizedError to provide a consistent experience for Salesforce responses.


lib/restforce.rb
```diff
@@ -42,7 +42,7 @@ module Tooling
   Error               = Class.new(StandardError)
   ServerError         = Class.new(Error)
   AuthenticationError = Class.new(Error)
-  UnauthorizedError   = Class.new(Error)
+  UnauthorizedError   = Class.new(Faraday::Error::ClientError)
```
timrogers: Why do you think this should inherit from a Faraday error when the others don't?

michaeldbianchi: I think the others should probably derive from ClientError as well, but this is the only Error type referenced in `lib/restforce/middleware/raise_error.rb` that does not have support for the response attribute. I'm honestly not sure when any of the other errors would really be used since they are not referenced in the `raise_error` middleware.

I'm specifically looking at the following:
https://github.com/restforce/restforce/blob/c79827f5ceb37fe32f21b21652c9a339bf733abb/lib/restforce/middleware/raise_error.rb#L19-L21

timrogers: Makes sense!

timrogers:
@michaeldbianchi Could you rebase this? Thanks!

PR #305, opened 2017-05-12
Support for JWT Bearer Token Authentication flow

tagCincy:
This PR adds the ability to authenticate to Salesforce via JWT Bearer Token.

- adds JWTBearer authentication middleware
- adds jwt gem (https://github.com/jwt/ruby-jwt) to handle JWT token generation
- adds option for passing path to private key
- updates readme

ref: https://github.com/ejholmes/restforce/issues/298

PR #479, opened 2019-09-12
Add jwt authentication flow support

nathanKramer:
Adds support for JWT authentication, based off of [this PR](https://github.com/restforce/restforce/pull/305/files)

@timrogers this works from my testing.


nathanKramer:
Hi @timrogers! Any chance you could take a look at this?

lib/restforce/middleware/authentication/jwt_bearer.rb
```diff
@@ -0,0 +1,34 @@
...
+      JWT.encode claim_set, private_key, 'RS256'
+    end
+
+    def claim_set
+      {
+          iss: @options[:client_id],
+          sub: @options[:username],
+          aud: @options[:host],
+          iat: Time.now.utc.to_i,
+          exp: 3.minutes.from_now.utc.to_i
```
timrogers: This doesn't work as ActiveSupport, which provides `Integer#minutes`, isn't a dependency of Restforce (and I'd prefer not to add it). Can you do this another way using pure Ruby?

nathanKramer: Good catch!
I've fixed that up, and this time I tested it with `script/console` rather than a rails console that required the gem 😄 

https://github.com/restforce/restforce/pull/479/files#diff-18cd1d533012fd67a233c8205c222769R26

lib/restforce/middleware/authentication/jwt_bearer.rb
```diff
@@ -0,0 +1,34 @@
...
+      JWT.encode claim_set, private_key, 'RS256'
+    end
+
+    def claim_set
+      {
+          iss: @options[:client_id],
+          sub: @options[:username],
+          aud: @options[:host],
+          iat: Time.now.utc.to_i,
+          exp: Time.now.utc.to_i + 180
```
nathanKramer: 👍

lib/restforce/middleware/authentication/jwt_bearer.rb
```diff
@@ -3,32 +3,36 @@
...
-  class Middleware::Authentication::JWTBearer < Restforce::Middleware::Authentication
-    def params
-      {
-        grant_type: 'urn:ietf:params:oauth:grant-type:jwt-bearer',
-        assertion: jwt_bearer_token
-      }
-    end
+  class Middleware
+    class Authentication
+      class JWTBearer < Restforce::Middleware::Authentication
```
nathanKramer: Not sure how I feel about this, the gem tends to prefer the compact style of `Middleware::Authentication::JWTBearer`, let me know if you'd prefer we tweak the rubocop config?

Would also be good to get the rubocop failures to fail PR checks on github

PR #452, opened 2019-05-07
Add support for platform events, CDC, generic events, etc

nathanKramer:
Unfortunately the streaming concern currently hardcodes `/topic/` in the channel name, which limits the number of use-cases it can be used for, especially now that Salesforce have [a number of other ways](https://trineo.com/blog/2019/04/push-vs-cdc) to publish cometD messages.

PR #492, opened 2019-10-21
Add support for platform events, CDC, generic events, etc

nathanKramer:
I'm not 100% sure what happened here, I saw that https://github.com/restforce/restforce/pull/452 was merged, however it seems to have been nuked from orbit so to speak, it's not in the master branch, as you can see here: https://github.com/restforce/restforce/blob/master/lib/restforce/concerns/streaming.rb (no sign of it in the commit history of the file)

I wonder if a certain jedi has being using the `--force`? 😆

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,8 +528,10 @@ client.get('/services/apexrest/FieldCase', company: 'GenePoint')
 
 ### Streaming
 
-Restforce supports the [Streaming API](http://wiki.developerforce.com/page/Getting_Started_with_the_Force.com_Streaming_API), and makes implementing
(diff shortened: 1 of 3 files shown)


GregPK:
Yes, noticed this as well with the same diagnosis.

timrogers:
Sorry about that! There was some Git fiddling required for the release and presumably this commit was lost somewhere on the way.

I will merge and cut a new release today ❤️

--

Tim Rogers
@timrogers

timrogers:
Not sure why Circle refuses to show in GitHub...

PR #495, opened 2019-11-03
Fix rspec warning during test runs

schmidt:
When running the test suite, rspec spits out the following warning:

```
`raise_error` was called with non-proc object nil
```

This is happening whenever `raise_error` is used with a non-block subject, because then, the code is executed before the expectation was set up and therefore it may yield unexpected results.

This change wraps the subject in a `lambda` so that the warning goes away and the spec works as expected. I also refactored a related spec, because due to the improved subject set up it may now become a bit simpler.

### DIFF
--- a/spec/unit/sobject_spec.rb
+++ b/spec/unit/sobject_spec.rb
@@ -45,13 +45,10 @@ describe Restforce::SObject do
     destroy: :destroy,
     destroy!: :destroy! }.each do |method, receiver|
     describe ".#{method}" do
-      subject(:send_method) { sobject.send(method) }
+      subject(:send_method) { lambda { sobject.send(method) } }
 
       context 'when an Id was not queried' do
-        it "raises an error" do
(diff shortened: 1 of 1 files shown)


timrogers:
Thanks! If you have any ideas on how I can persuade GitHub to show the CircleCI runs for this project, I'd love to hear them.

schmidt:
>  If you have any ideas on how I can persuade GitHub to show the CircleCI runs for this project, I'd love to hear them.

Not sure. I've logged into CircleCI, added schmidt/restforce as a project and the next PR had these fancy checks.

https://github.com/schmidt/restforce/pull/1

In the project's settings I see, that CircleCI set up a deploy key and a webhook. 

Did you check the "Advanced Settings" at https://circleci.com/gh/restforce/restforce/edit#advanced-settings ? There's a setting to also "Build forked pull requests". Maybe that one is missing?

PR #499, opened 2019-11-15
Add support for sObject Collections (4.2.0)

spectator:
Based on recent master which is gem v4.2.0


spectator:
Sorry, wrong target 🤦‍♂

PR #503, opened 2019-12-12
Fix nil attributes error

apurkiss:
This fixes an issue in mash when the `attributes` key is present but the value is nil. Currently when this happens it fails on line 31: `case (val['attributes']['type'])` with:
```
NoMethodError: undefined method `[]' for nil:NilClass
```
Now when this happens we just return `Restforce::SObject`.

### DIFF
--- a/lib/restforce/mash.rb
+++ b/lib/restforce/mash.rb
@@ -28,6 +28,8 @@ module Restforce
           # of sobject records.
           Restforce::Collection
         elsif val.key? 'attributes'
+          return Restforce::SObject unless val['attributes']
+
           case (val['attributes']['type'])
           when "Attachment"
             Restforce::Attachment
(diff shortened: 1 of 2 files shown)


lib/restforce/mash.rb
```diff
@@ -28,6 +28,8 @@ def klass(val)
           # of sobject records.
           Restforce::Collection
         elsif val.key? 'attributes'
+          return Restforce::SObject unless val['attributes']
+
           case (val['attributes']['type'])
```
gfmurphy: @apurkiss just a suggestion: i think you can change this line to `val.dig('attributes', 'type')` and remove the guard clause, which would be slightly less complex maybe?

apurkiss:
@ejholmes / @timrogers any chance we can get this reviewed? Thanks in advance for your time!

jeffdonthemic:
@timrogers Looking for these changes as well.

timrogers:
Sorry for the delay!

apurkiss:
Thanks @timrogers!

PR #507, opened 2019-12-27
Clarify that refresh process uses host option

drewish:
I spent *way* longer than I'd like to admit trying to figure out why my refresh tokens weren't working correctly. Eventually I discovered they were being sent to the production environment rather than the sandbox one. Re-reading the docs I see that mentioned but since it appeared under the proxy section I'd over looked it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Restforce is a ruby gem for the [Salesforce REST api](http://www.salesforce.com/
 Features include:
 
 * A clean and modular architecture using [Faraday middleware](https://github.com/technoweenie/faraday) and [Hashie::Mash](https://github.com/intridea/hashie/tree/v1.2.0)'d responses.
-* Support for interacting with multiple users from different orgs.
+* Support for interacting with multiple users from different organizations.
 * Support for parent-to-child relationships.
 * Support for aggregate queries.
 * Support for the [Streaming API](#streaming)
@@ -48,7 +48,7 @@ so you can do things like `client.query('select Id, (select Name from Children__
 ### Initialization
 
 Which authentication method you use really depends on your use case. If you're
-building an application where many users from different orgs are authenticated
+building an application where many users from different organizations are authenticated
 through oauth and you need to interact with data in their org on their behalf,
 you should use the OAuth token authentication method.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,7 +142,17 @@ export SALESFORCE_API_VERSION="41.0"
 client = Restforce.new
 ```
 
-### Proxy Support
+#### Sandbox Orgs
```
drewish: I moved this up since my guess is more people test in sandbox than use proxies.

README.md
```diff
@@ -142,7 +142,17 @@ export SALESFORCE_API_VERSION="41.0"
...
+
+You can connect to sandbox orgs by specifying a host. The default host is
+'login.salesforce.com':
+
+```ruby
+client = Restforce.new(host: 'test.salesforce.com')
+```
+The host can also be set with the environment variable `SALESFORCE_HOST`.
+
+#### Proxy Support
```
drewish: I changed this from an h3 to h4 since it seemed on the same level as the other option documentatation.

README.md
```diff
@@ -78,7 +78,7 @@ client = Restforce.new(oauth_token: 'access_token',
                        api_version: '41.0')
 ```
 
-The middleware will use the `refresh_token` automatically to acquire a new `access_token` if the existing `access_token` is invalid.
+The middleware will use the `refresh_token` automatically to acquire a new `access_token` if the existing `access_token` is invalid. The refresh process uses the `host` option so make sure that is set correctly for sandbox orgs.
```
timrogers, line 81: Can we use `organisations` instead of `orgs`?

README.md
timrogers, line 145: See above ⏫

README.md
```diff
@@ -142,7 +142,17 @@ export SALESFORCE_API_VERSION="41.0"
 client = Restforce.new
 ```
 
-### Proxy Support
+#### Sandbox Orgs
+
+You can connect to sandbox orgs by specifying a host. The default host is
```
timrogers, line 147: See above ⏫

timrogers:
Sorry for the delay in getting to this PR - I didn't spot it!

drewish:
@timrogers want to give it another look?

PR #509, opened 2020-01-06
Fallback to building a Restforce::Mash if the attributes hash does not contain a type

jcn:
When trying to make arbitrary API calls to, for example, the `/ui-api/object-info/<object>/picklist-values/` endpoint, the response contains an attributes key but no type key, which was causing the following:

```
NoMethodError: undefined method `[]' for nil:NilClass
from .../restforce-4.2.1/lib/restforce/mash.rb:32:in `klass'
```

This patch ensures that there is a type key inside of the attributes hash, otherwise it just falls back to Restforce::Mash.

### DIFF
--- a/lib/restforce/mash.rb
+++ b/lib/restforce/mash.rb
@@ -27,7 +27,7 @@ module Restforce
           # When the hash has a records key, it should be considered a collection
           # of sobject records.
           Restforce::Collection
-        elsif val.key? 'attributes'
+        elsif val.dig('attributes', 'type')
           case (val['attributes']['type'])
           when "Attachment"
             Restforce::Attachment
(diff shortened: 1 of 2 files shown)


apurkiss:
Hey @jcn. I've got a [PR](https://github.com/restforce/restforce/pull/503) up for a very similar change. I was also thinking of defaulting to `Restforce::Mash`. However, in the case of querying for dashboards for example, there can be a `nil` attributes value yet this must be treated as an SObject. As the comments point out:

```
            # When the hash contains an attributes key, it should be considered an
            # sobject record
```

Also I've been waiting for a review for awhile. Not sure if the repo is actively being maintained. Maybe the holiday break is delaying things.

jcn:
@apurkiss Hey, I missed your PR - I'm happy to close mine and I'm not actually sure which is the right behavior (Mash or SObject). I'd leave it to the maintainers to decide, but it sure would be nice to get one of these merged in soon!

jcn:
Closed because #503 takes solves the problem.

Issue #501, opened 2019-11-28
without_caching doesn't seem to work


nduncan-cs130:
I noticed that the same cache entries were being created regardless of whether or not I wrapped the API call in a `client.without_caching` block

From what I could tell this is the code that would enforce the `without_caching` block:
```
  class Middleware::Caching < FaradayMiddleware::Caching
    def call(env)
      expire(cache_key(env)) unless use_cache?
      super
    end
....
``` 

It looks like it expires the cache entry, then calls super. But I'm assuming super will write the entry to the cache, so the expiration has no effect.

Changing the code to expire the cache entry after calling super produces the correct behavior:
```
 class Middleware::Caching < FaradayMiddleware::Caching
    def call(env)
      super
    ensure
      expire(cache_key(env)) unless use_cache?
    end
...
```

swaincreates:
The way things are setup (from my understanding of `Restforce::Concerns::Connection`) is that you cannot prevent writing to the cache, but this is a solution to expire it and make a new request so the results will be fresh every time the request is made within a `without_caching` block.

PR #516, opened 2020-01-30
Adds ability to opt in to caching on a per-request basis

swaincreates:
I would like to take advantage of Restforce's caching feature but would prefer to opt in as I make changes to the code base. This adds the ability to opt in to caching on a per-request basis similar to `.without_caching` method opts out on a per-request basis. 

The changes I made here seemed to require the least amount of systematic alterations. It adds a method and makes a private config public with a default. The one negative I see to this option is that because of the way things are setup in `Restorce::Concerns::Connection`, all requests will actually be cached. Only those within `.with_caching` will not expire the cache first, forcing a fresh request, thus accomplishing my intent.

It seems like if we wanted to only write to the cache in `.with_caching` or likewise not write to the cache in `.without_caching` as mentioned in https://github.com/restforce/restforce/issues/501 then we would need to build the Restforce::Concerns::Connection `@connection` each time instead of memoizing it, and have an option to include middleware in addition to `if cache`. See line 44 from Restforce::Concerns::Connection below:

```ruby
          builder.use Restforce::Middleware::Caching, cache, options if cache
```

There may be more to this than I'm seeing, but I thought I would go ahead and make this PR in case people thought it would be useful. I certainly understand if you think its out of scope and I'll go a different route.

### DIFF
--- a/README.md
+++ b/README.md
@@ -688,6 +688,22 @@ client.without_caching do
 end
 ```
 
+If you prefer to opt in to caching on a per-request, you can do so by using .with_caching and
+setting the use_cache config option to false:
+
+```ruby
+Restforce.configure do |config|
+  config.cache = Rails.cache
+  config.use_cache = false
+end
+```
+
+```ruby
+client.with_caching do
+  client.query('select Id from Account')
+end
+```
+
 Caching is done on based on your authentication credentials, so cached responses will not be shared between different Salesforce logins.
 
 * * *
--- a/README.md
+++ b/README.md
@@ -689,7 +689,7 @@ end
 ```
 
 If you prefer to opt in to caching on a per-request, you can do so by using .with_caching and
-setting the use_cache config option to false:
+setting the `use_cache` config option to false:
 
 ```ruby
 Restforce.configure do |config|
@@ -704,7 +704,7 @@ client.with_caching do
 end
 ```
 
-Caching is done on based on your authentication credentials, so cached responses will not be shared between different Salesforce logins.
+Caching is done based on your authentication credentials, so cached responses will not be shared between different Salesforce logins.
 
 * * *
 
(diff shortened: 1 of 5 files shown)


timrogers:
@thomasklemm @ddayporter Could one of you review this too? <3

README.md
```diff
@@ -688,6 +688,22 @@ client.without_caching do
 end
 ```
```
timrogers: `Client#with_caching`

README.md
```diff
@@ -688,6 +688,22 @@ client.without_caching do
...
 
+If you prefer to opt in to caching on a per-request, you can do so by using .with_caching and
+setting the use_cache config option to false:
```
timrogers: `use_cache`

timrogers:
@swaincreates Nice work! Just a couple of tiny suggestions for me.

Have you had a chance to test this in production?

README.md
```diff
@@ -688,6 +688,22 @@ client.without_caching do
...
+end
+```
+
+```ruby
+client.with_caching do
+  client.query('select Id from Account')
+end
+```
+
 Caching is done on based on your authentication credentials, so cached responses will not be shared between different Salesforce logins.
```
ddayporter: Looks like an extra 'on' in this sentence - may as well fix while you're at it? 😄

README.md
swaincreates: Sorry, maybe it's the Github UI, but I'm not sure what you're asking for here. I tried to follow the same kind of formatting used in the README and I dont think I changed anything in this part. Can you clarify?

README.md
swaincreates: Once again, not 100% sure what you're recommending - you just want me to just add backticks around use_cache?

timrogers: Exactly!

README.md
swaincreates: I made an update based on your other comments, but I cant figure this one out - I think this section is from the current README.md

swaincreates:
Coming back to this - I updated the README.md in a couple of places mentioned but I had one suggestion that I couldn't quite figure out as mentioned above.

PR #528, opened 2020-04-17
Add create, update, destroy, and upsert for Composite API.

meenie:
This adds in the `composite` and `composite!` methods to work with the Composite API located here: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_composite.htm.  The `batch` method did not work for us because each sub-request still counts as an individual API call.  When using the Composite API all 25 of the sub-requests will count as 1 API call.  I have no idea why Salesforce decided to do that...but...here we are.

As you can no doubt tell, I have based this code addition entirely on the previously written `BatchAPI` concern with a few minor tweaks.  For one, you can't break up multiple sub-requests into groups of 25 because the Composite API makes use of Reference ID's to allow a developer to reference previous sub-requests.  So I've put a hard limit of 25 sub-requests and the developer will have to do their own batching.

We currently don't have a use for the `GET` feature of the Composite API so it hasn't been implemented.


amacdougall:
Hello! I'm taking on responsibility for this pull request—is there anything I can do to help it pass review? This change looks quite self-contained, so I don't believe it has any risk of effects beyond its own domain.

piotrjaworski:
Amazing work @amacdougall , looks really good!

Any possilbe update here? @meenie

balabis:
Any updates on this?

timrogers:
@amacdougall Your feedback that this is working for you in production gives me the confidence I couldn't have alone to merge this 😹 Would you be able to write me a quick blurb on what the Composite API is? And perhaps make some appropriate changes to the readme? (cc @meenie)

amacdougall:
Sure thing, I'll add that information soon! The quick version, though: the Composite API makes it possible to work with object graphs rather than just individual records. In our use case, for instance, we have custom Survey, Response, and Answer objects, each nested. With the standard API, we would have to create a Response in one request, then create its Answers in further requests. With the Composite API, we can send that entire nested structure in a single request.

Our system used to use the Bulk API, but that system is designed for infrequent uploads of _mass_ data, not frequent uploads of a handful at a time, so it has much lower API limits. Using the Composite API for this purpose has really saved our bacon. And if you compare the code, you'll see that this Composite API implementation is very similar to the Bulk API code already in the gem (though I have to double-check this assertion).

joegaudet:
Thanks all, very excited to use this.

We’ve been working on a new ActiveModel adapter we will likely open source
once this work is in.

The implications for auto save are going to be great!

.joe

timrogers:
@amacdougall Sorry to pull you back to this old PR, but @robob27 has flagged in #679 that this code doesn't seem to be working for him, because the responses he gets from the Salesforce API have different casing in the JSON keys (`compositeResponse` vs `CompositeResponse`) to what the code expects.

Can you confirm that you're using v5.2.0 in production? If so, which Salesforce API version are you using?

timrogers:
cc @meenie on the above too 👋

PR #531, opened 2020-04-22
Update faraday and faraday_middleware dependencies

flaviovisetti:
Update of `faraday` and `faraday_middleware` dependencies.
With this update I fixed the specs that broke.

These updates are very important as many other gems do not work alongside restforce.

Ex: `elasticsearch`, '~> 7.6.0'

### DIFF
--- a/lib/restforce/upload_io.rb
+++ b/lib/restforce/upload_io.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Restforce
-  UploadIO = Faraday::UploadIO
+  UploadIO = Faraday::FilePart
 end
 
 require 'restforce/patches/parts' unless Parts::Part.method(:new).arity.abs == 4
--- a/lib/restforce/upload_io.rb
+++ b/lib/restforce/upload_io.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'faraday/upload_io'
-
 module Restforce
   UploadIO = Faraday::UploadIO
 end
--- a/spec/unit/concerns/connection_spec.rb
+++ b/spec/unit/concerns/connection_spec.rb
@@ -75,7 +75,7 @@ describe Restforce::Concerns::Connection do
 
       it "must always be used last before the Faraday Adapter" do
         client.middleware.handlers.reverse.index(Restforce::Middleware::Logger).
-          should eq 1
+          should eq 0
       end
     end
   end
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 gemspec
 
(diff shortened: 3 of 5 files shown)


flaviovisetti:
@timrogers Could you check this PR?

spec/unit/concerns/connection_spec.rb
```diff
@@ -75,7 +75,7 @@
 
       it "must always be used last before the Faraday Adapter" do
         client.middleware.handlers.reverse.index(Restforce::Middleware::Logger).
-          should eq 1
+          should eq 0
```
phrsouza, line 78: I'm not sure about this spec. @timrogers Can you take a look?

lib/restforce/upload_io.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'faraday/upload_io'
-
 module Restforce
   UploadIO = Faraday::UploadIO
```
timrogers, line 4: Does this still work? Is it just the require that is no longer needed?

spec/unit/concerns/connection_spec.rb
timrogers: This suggests that the Faraday adapter is missing - which could be a problem!

flaviovisetti: @timrogers Handles and Adapters are separated now. 

`https://github.com/lostisland/faraday/blob/master/UPGRADING.md#others-1`

lib/restforce/upload_io.rb
flaviovisetti: For a better understanding, I am updated the method to the new name of UploadIO
`Faraday::FilePart`

spec/unit/concerns/connection_spec.rb
flaviovisetti: In the connection file the builder already separates by pointing to the adapter. 

https://github.com/restforce/restforce/blob/master/lib/restforce/concerns/connection.rb#L64

Do you think this Faraday modification is a problem?

lib/restforce/upload_io.rb
timrogers: I’d be concerned that this might be incompatible with older Faraday versions.

flaviovisetti: Could be in the future. According this faraday PR the method `UploadIO` is deprecated but functional.

I think they will remove that in the future.

https://github.com/lostisland/faraday/pull/1021/files

timrogers:
> Hey @timrogers, anything blocking the review of this PR?

👋 Only my busyness.

Have you been able to test this? I do t have access to a Salesforce instance and I’m sorry that some changes might cause issues here.

timrogers:
Done with another PR!

PR #546, opened 2020-06-25
Add support for faraday 1.x

### LINKED ISSUE #530: faraday and faraday_middleware are outdated
Any ideia when this PR will be merge and create release?
Without this update many other gems does not work with Restforce. 

ex: `gem 'elasticsearch', '~> 7.6.0'`

Notice that both gems needs to be update, Faraday and Faraday Middleware.

https://github.com/restforce/restforce/pull/526

```
  gem.add_dependency 'faraday', '>= 0.9.0', '<= 2.0'
  gem.add_dependency 'faraday_middleware', ['>= 0.8.8', '<= 2.0']
```

ryansch:
I followed: https://github.com/lostisland/faraday/blob/master/UPGRADING.md

I also looked at the differences between older versions (https://github.com/lostisland/faraday/blob/v0.15.4/lib/faraday/error.rb) and the latest version (https://github.com/lostisland/faraday/blob/v1.0.1/lib/faraday/error.rb) to see if there was common behaviour we could exploit. That's where rescuing `Faraday::Error` came from.

Fixes #530
Closes #526
Closes #544

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@
 source 'https://rubygems.org'
 gemspec
 
-gem 'faraday', '~> 0.17.0'
 gem 'jruby-openssl', platforms: :jruby
 gem 'jwt'
 gem 'rake'
--- a/README.md
+++ b/README.md
@@ -457,7 +457,7 @@ Using the new [Blob Data](http://www.salesforce.com/us/developer/docs/api_rest/C
 client.create('Document', FolderId: '00lE0000000FJ6H',
                           Description: 'Document test',
                           Name: 'My image',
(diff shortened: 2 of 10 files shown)


ryansch:
I didn't see #544. I'll update this with those changes as well.

morganick:
🎉

timmccormick:
🎉

timrogers:
Thanks for this, and sorry for the slow review! I'm going to somewhat trust that this one works and release it as a release candidate and it's a bit of a scary one. If people upgrade successfully and don't have issues, then we can release it as a final version.

PR #550, opened 2020-07-08
Explicitly define classes for error codes

presidentbeef:
There are a _lot_ of different ways to do this, so please suggest away on changes. I am happy to adjust.

For example:

- Do error definitions with `Class.new` instead. Can align text on `=`
- Still create errors dynamically, but from a fixed list
- Raise a different exception/message for missing error codes
- Put `error_codes.rb` somewhere else
- ...etc.

Thanks for having a nice clear code and tests! :+1:


lib/restforce.rb
```diff
@@ -60,9 +60,11 @@ module Tooling
   MatchesMultipleError= Class.new(ResponseError)
   EntityTooLargeError = Class.new(ResponseError)
 
+  require 'restforce/error_codes'
+
   module ErrorCode
     def self.const_missing(constant_name)
-      const_set constant_name, Class.new(ResponseError)
+      raise ArgumentError, "No error code defined for #{constant_name}"
```
timrogers, line 67: On second thoughts, I think we should take a slightly more risk-averse approach here and continue to define constants which are not found, but log some kind of warning saying that an unexpected error was encountered and that the user should report it to us at https://github.com/restforce/restforce/issues/new/choose.

That will ensure that we don't accidentally break someone's code. In a future version, once we are more confident that we have all of the errors defined, we could remove this.

lib/restforce/error_codes.rb
```diff
@@ -0,0 +1,184 @@
+# frozen_string_literal: true
+
+# Define error codes based on
+# https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_concepts_core_data_objects.htm#statuscode
+module Restforce
+  module ErrorCode
+    class AllOrNoneOperationRolledBack < ResponseError; end
```
timrogers, line 7: I would be in favour of adding some kind of "lookup table" here (e.g. a hash which maps the `errorCode` to the class) to get rid of the `const_get` weirdness.

timrogers: I would make that hash lookup "fall back" to `const_get` if it fails.

lib/restforce.rb
presidentbeef: There are two scenarios where code might break:

1. They reference `Restcode::ErrorCode::Something` in their code that isn't defined.
2. The API returns an error code not mapped to an existing constant.

I think (1) is not a problem - the failure will be obvious as soon as the library is updated.

(2) can be addressed by removing the reliance on `const_get`/`const_missing` in `exception_class_for_error_code` and just defining the constant there.

timrogers: (1) is true assuming that they are not doing some kind of weird const lookup - which is probs it a fair assumption to make. I think we should address (2) by doing a hash lookup (as I describe in my other comment), and if the hash lookup fails:

1. Define the constant
2. Log a warning asking the user to create an issue reporting the missing error

presidentbeef:
@timrogers I have moved the logic for fetching/creating new error code classes into `lib/restforce/error_codes.rb`.

I believe this covers your previous comments.

Once again, totally open to any changes you would like me to make.

By the way - might be nice to upgrade RSpec to get more nice matchers like `output_to_stderr` :smiley:

timrogers:
Closed in favour of #551 - thanks for your work on this!

PR #551, opened 2020-07-09
Explicitly define error classes for known Salesforce errors

timrogers:
The Salesforce REST API can return a range of `errorCode`s representing different kinds of errors. To make these easy to
handle in your code, we want to turn these into individual, specific exception classes in the `Restforce::ErrorCode` namespace that inherit from `Restforce:: ResponseError`.

Today, this behaviour is achieved by taking the error code returned by Salesforce (e.g. `ALL_OR_NONE_OPERATION_ROLLED_BACK`), normalising it into something that looks like a Ruby class name (e.g. `AllOrNoneOperationRolledBack`) and then looking it up inside the `Restforce::ErrorCode` module using `const_get` to find the correct error class to raise.

`Restforce::ErrorCode` then uses `const_missing` to dynamically define the error class requested - so in short, if Salesforce returns an error, a class will be defined for it.

Why change this? A few reasons:

1. It makes it easy to accidentally write wrong code. I could write `rescue Restforce::ErrorCode::ErrorWithWrongSpeling` and this code would be valid, as `const_missing` would define the constant for me. But I wouldn't be rescuing what I thought I was. (The same goes for referring to an error that it totally wrong - it is not just spelling mistakes.)
2. The library doesn't have any clear list anywhere of the errors it *might* raise, and it is impossible to generate such a list. This doesn't make the library terribly easy to work with.
3. The dynamic definition of constants raises the spectre of exploits of users' code which might allow them to take up lots of the user's memory by creating many exception classes. This would only be an issue, of course, if you were doing some kind of dynamic lookup yourself.

This code changes the approach so we take a list of known error codes from the Salesforce documentation  and define the error classes for them upfront.

The bad part of this change is that we have to decide what to do when we get an error back from Salesforce that isn't predefined.

I've decided to go for raising a generic `Restforce::ResponseError` in this case, and logging a warning telling the user to report the missing error on GitHub. __This has the potential of being a breaking change as there could be real Salesforce errors that we don't know about, which users of Restforce will no longer be able to rescue (as they don't exist) and which will raise a new generic `ResponseError`).

We will also log a warning if someone refers to an class in the `ErrorCode` namespace that doesn't exist.

Thanks to @presidentbeef for his initial work on this in #550.

### DIFF
--- a/lib/restforce/error_code.rb
+++ b/lib/restforce/error_code.rb
@@ -1,0 +1,406 @@
+# frozen_string_literal: true
+
+module Restforce
+  module ErrorCode
+    GITHUB_ISSUE_URL = "https://github.com/restforce/restforce/issues/new?template=" \
+      "unhandled-salesforce-error.md&title=Unhandled+Salesforce+error%3A+%3Cinsert+" \
+      "error+code+here%3E"
+
+    # We define all of the known errors returned by Salesforce based on the
+    # documentation at
+    # https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_concepts_core_data_objects.htm#statuscode
+    # Previously, these were defined dynamically at runtime using `Module.const_set`. Now
+    # we define them up-front.
+    # It is possible that we will be missing some errors, so we will handle this in
+    # at least a semi-graceful manner.
+    class AllOrNoneOperationRolledBack < ResponseError; end
+    class AlreadyInProcess < ResponseError; end
+    class AssigneeTypeRequired < ResponseError; end
+    class BadCustomEntityParentDomain < ResponseError; end
+    class BccNotAllowedIfBccComplianceEnabled < ResponseError; end
+    class BccSelfNotAllowedIfBccComplianceEnabled < ResponseError; end
+    class CannotCascadeProductActive < ResponseError; end
+    class CannotChangeFieldTypeOfApexReferencedField < ResponseError; end
+    class CannotCreateAnotherManagedPackage < ResponseError; end
+    class CannotDeactivateDivision < ResponseError; end
+    class CannotDeleteLastDatedConversionRate < ResponseError; end
+    class CannotDeleteManagedObject < ResponseError; end
+    class CannotDisableLastAdmin < ResponseError; end
+    class CannotEnableIpRestrictRequests < ResponseError; end
+    class CannotInsertUpdateActivateEntity < ResponseError; end
+    class CannotModifyManagedObject < ResponseError; end
+    class CannotRenameApexReferencedField < ResponseError; end
+    class CannotRenameApexReferencedObject < ResponseError; end
+    class CannotReparentRecord < ResponseError; end
+    class CannotResolveName < ResponseError; end
+    class CannotUpdateConvertedLead < ResponseError; end
+    class CantDisableCorpCurrency < ResponseError; end
+    class CantUnsetCorpCurrency < ResponseError; end
+    class ChildShareFailsParent < ResponseError; end
+    class CircularDependency < ResponseError; end
+    class CommunityNotAccessible < ResponseError; end
+    class CustomClobFieldLimitExceeded < ResponseError; end
+    class CustomEntityOrFieldLimit < ResponseError; end
+    class CustomFieldIndexLimitExceeded < ResponseError; end
+    class CustomIndexExists < ResponseError; end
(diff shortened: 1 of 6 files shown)


timrogers:
@nickcampbell18 Hey old friend 👋  Would you have some time to look at this? I'd be interested to hear your thoughts on the change and how you would version it.

lib/restforce/error_code.rb
```diff
@@ -0,0 +1,408 @@
...
+    class UnspecifiedEmailAddress < ResponseError; end
+    class UnsupportedApexTriggerOperation < ResponseError; end
+    class UnverifiedSenderAddress < ResponseError; end
+    class WeblinkSizeLimitExceeded < ResponseError; end
+    class WeblinkUrlInvalid < ResponseError; end
+    class WrongControllerType < ResponseError; end
+
+    # Maps `errorCode`s returned from Salesforce to the exception class
+    # to be used for these errors
+    ERROR_EXCEPTION_CLASSES = {
```
nickcampbell18, line 195: I'm worried about how you keep these two lists synchronised (e.g. defining the error class without the mapping). Unfortunately the alternative looks a bit meta-programmy right?

Throwing line lengths to the wind....

```ruby
[
  "ALL_OR_NONE_OPERATION_ROLLED_BACK" => AllOrNoneOperationRolledBack = Class.new(ResponseError)
]
```

Alternative, you could apply the registration pattern. IDEs and language servers will still recognise the classes being registered here, so you can just meta-program the error hash that you want to build:

```ruby
class ResponseError
  def self.inherited(klass)
    ERROR_CLASSES[klass.snake_case.upcase] = klass
  end
end

class StringTooLong < ResponseError; end

ERROR_CLASSES 
#=> { "STRING_TOO_LONG" => StringTooLong }
```

timrogers:
What would you think of adding some specs which attempt to make sure that all of the errors are “mapped”? You could potentially use `#constants` to find all of the defined error classes and then make sure they are “mapped”.

I like that as I would rather have meta programming in the tests than during runtime.

lib/restforce/error_code.rb
timrogers: What would you think of adding some specs which attempt to make sure that all of the errors are “mapped”? You could potentially use `#constants` to find all of the defined error classes and then make sure they are “mapped” in the hash. 

I like that as I would rather have meta programming in the tests than during runtime.

lib/restforce/error_code.rb
```diff
@@ -0,0 +1,408 @@
...
+      "UNSPECIFIED_EMAIL_ADDRESS" => UnspecifiedEmailAddress,
+      "UNSUPPORTED_APEX_TRIGGER_OPERATION" => UnsupportedApexTriggerOperation,
+      "UNVERIFIED_SENDER_ADDRESS" => UnverifiedSenderAddress,
+      "WEBLINK_SIZE_LIMIT_EXCEEDED" => WeblinkSizeLimitExceeded,
+      "WEBLINK_URL_INVALID" => WeblinkUrlInvalid,
+      "WRONG_CONTROLLER_TYPE" => WrongControllerType
+    }.freeze
+
+    def self.get_exception_class(error_code)
+      error_class = ERROR_EXCEPTION_CLASSES[error_code]
```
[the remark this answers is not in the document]
timrogers: I tried that and it didn’t seem to work in Ruby 2.4. I have dropped support for it now so I can probably switch back to this more elegant solution.

lib/restforce/error_code.rb
timrogers: If you What would you think of adding some specs which attempt to make sure that all of the errors are “mapped”? You could potentially use `#constants` to find all of the defined error classes and then make sure they are “mapped”.

I like that as I would rather have meta programming in the tests than during runtime. like that idea, I’d also be cool with number 1. I’d rather avoid mutating the constant as in number 2 it feels like we’re replacing one weird behaviour with another.

nickcampbell18: Also using specs works!

lib/restforce/error_code.rb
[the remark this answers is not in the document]
timrogers: Working fine in Ruby 2.5+! Thanks for the push on this.

lib/restforce/error_code.rb
timrogers: This is now enforced using tests.

Issue #566, opened 2020-08-27
v4.x gemspec blocks important Faraday fixes (e.g. can result in zombie TCP connections)


hongtron:
the v4.2.2 gemspec specifies [`gem.add_dependency 'faraday', '<= 1.0', '>= 0.9.0'`](https://github.com/restforce/restforce/blob/v4.2.2/restforce.gemspec#L28), which blocks patch releases of Faraday 1.0. Faraday 1.0.1 has an important fix to prevent TCP connections from being left open: https://github.com/lostisland/faraday/pull/1117

Our ideal solution would be a 4.2.3 release that specifies `'< 1.1'` instead of `'<= 1.0'`. This would unblock our gem version upgrades, which would unblock our upgrade from Ruby 2.4, which would unblock us from upgrading to v5.x of `restforce` 🙂

timrogers:
@hongtron Thanks for reporting this! Can you create a PR for the change?

hongtron:
@timrogers definitely! However, I'm not seeing a v4 branch that I can make a PR against. Do you have a preferred process for pushing updates to past major versions? I don't see any guidelines in `CONTRIBUTING.md`.

hongtron:
@timrogers bump ^^ 🙏🏼

hongtron:
Here's the commit that I'm not able to merge in: https://github.com/hongtron/restforce/commit/d53e88d044cf48ced8d6557225ebad1de774c9c0

hongtron:
Closed via https://github.com/restforce/restforce/pull/622.

PR #618, opened 2021-01-18
Allow Faraday versions up to, but not including, v1.1 to be used with Restforce 4.x

timrogers:
This allows Faraday versions up to, but not including, `v1.1` to be used with Restforce.

_This pull request merges into the `v4.x` branch rather than the current mainline `master` branch._

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@
 source 'https://rubygems.org'
 gemspec
 
-gem 'faraday', '~> 0.17.0'
+gem 'faraday', '~> 1.0.1'
 gem 'jruby-openssl', platforms: :jruby
 gem 'jwt'
 gem 'rake'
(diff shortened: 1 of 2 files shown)


timrogers:
Addresses #566.

@hongtron Any ideas on how to fix the errors here?

hongtron:
Looks like [this rename](https://github.com/lostisland/faraday/pull/1021), which was part of the [Faraday 1.0.0 release](https://github.com/lostisland/faraday/releases/tag/v1.0.0), is the culprit.

If Restforce is going to continue to support Faraday versions from before and after this rename, then you may want to check the Faraday version before requiring [here](https://github.com/restforce/restforce/blob/9ba34209eadc7115a75f04cbe01b36dd3b9599b7/lib/restforce/upload_io.rb):

```ruby
if Faraday.VERSION =~ /0\./
  require 'faraday/upload_io'
else
  require 'faraday/file_part'
end
```

That's just my first instinct, though - I didn't dig into the patching behavior that's being applied there.

timrogers:
Closing in favour of alternative approach now in release candidate

Issue #627, opened 2021-02-19
Streaming subscription:No way to get the channel if you subscribe to more than one.


KludgeKML:
I have a situation in which I need to subscribe to events from more than one channel. I see I can do that like this:

KludgeKML:
Related PR:
https://github.com/restforce/restforce/pull/628

PR #628, opened 2021-02-19
Allow passing an option to client.subscription which will return both channel and message in the callback block

KludgeKML:
Currently if you want to listen to events from more than one channel, you can pass an array of channels to the client.subscription call:

```
EM.run do
  # Subscribe to multiple PushTopics.
  client.subscription ['/topic/AllAccounts', '/topic/AllUsers'] do |message|
    puts message.inspect
  end
end
```

The downside to this approach is that you will not necessarily know which channel the message came from. This PR allows you to add a `with_channel:` option to the subscription call. If true, the callback will provide the channel and message:

```
EM.run do
  # Subscribe to multiple PushTopics.
  client.subscription(['/topic/AllAccounts', '/topic/AllUsers'], with_channel: true) do |channel, message|
    puts "Message came in on channel #{channel}"
    puts message.inspect
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -572,6 +572,32 @@ end
 Boom, you're now receiving push notifications when Accounts are
 created/updated.
 
+#### Receiving events from more than one channel
+
+If you want to listen to events from more than one channel, you can pass an array of channels to the client.subscription call:
+
+```
+EM.run do
+  # Subscribe to multiple PushTopics.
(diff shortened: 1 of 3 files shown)


KludgeKML:
Related issue: 
https://github.com/restforce/restforce/issues/627

PR #668, opened 2021-09-29
Add fast path for block-less Restforce::Collection#count

jhass:
Calling `count` is an easy footgun currently as it needlessly loads the entire collection.

### DIFF
--- a/lib/restforce/collection.rb
+++ b/lib/restforce/collection.rb
@@ -32,6 +32,12 @@ module Restforce
       @raw_page['totalSize']
     end
     alias length size
+    
+    def count
+      return size unless block_given?
+      
+      super
+    end
 
     # Returns true if the size of the Collection is zero.
     def empty?


lib/restforce/collection.rb
```diff
@@ -33,6 +33,12 @@ def size
     end
     alias length size
 
+    def count
```
timrogers, line 36: Do we need to do something to handle the `count(item) -> int` 
case? See the Ruby standard library docs [here](https://ruby-doc.org/core-3.0.2/Enumerable.html#method-i-count).

timrogers: I'm wondering if it should be `def count(item = nil)`.

jhass: Yes, good catch!

`item = nil` wouldn't work, it would count how many items would equal to `nil`. Declaring it to take any number of parameters works as `super` forwards the arg list as is it then allows the C land to inspect it. Will look a bit ugly in the docs though :/

timrogers: I don't write much Ruby anymore so I'm glad I've still got it in me - I did slightly fear my feedback could be totally off base 😅

lib/restforce/collection.rb
```diff
@@ -33,6 +33,12 @@ def size
     end
     alias length size
 
+    def count(*args)
+      return size unless block_given? || !args.empty?
```
timrogers, line 37: What do you think of adding a comment to explain the "why" behind this implementation - why we need this condition and why we bother to delegate to `#size`? I suspect it could cause some future person to wonder what's going on.

PR #679, opened 2021-11-18
Fix compositeResponse capitalization

### LINKED ISSUE #678: Mismatch in capitalization of string keys in composite API response causes errors
Rest API Version: 53
Restforce Version: 5.2.0
Organization Edition: Unlimited Edition
Organization Release: Winter '22
Organization Type: Sandbox

👋 I'm attempting to use the Composite API through Restforce for the first time. I submitted a request to create an Account with an associated Contact like so:

```
client = Salesforce.rest_client # creates a restforce client

requests = []

requests << {
	method: :create,
	sobject: "Account",
	reference_id: "refAccount",
	data: {
		Name: "Rob Test Composite Account",
		Website: "robtestcompositeaccount.com"
	}
}

requests << {
	method: :create,
	sobject: "Contact",
	reference_id: "refContactId",
	data: {
		LastName: "RobTestCompositeContact",
		AccountId: "@{refAccount.id}"
	}
}

requests.each_slice(25).map do |req_slice|
  client.composite do |subrequest|
    req_slice.each do |r|
      subrequest.send *r.values
    end
  end
end
```

Initially when I tried this I received the following error:

```
NoMethodError: undefined method `any?' for nil:NilClass
from /Users/robgoodberry/.gem/ruby/2.7.1/gems/restforce-5.2.0/lib/restforce/concerns/composite_api.rb:32:in `composite'
```

But I noticed that the Account and Contact were actually successfully created in Salesforce. I used pry to check the actual response that was coming back which looked like:

```
   response_body=
    {"compositeResponse"=>
      [{"body"=>{"id"=>"0011k00000wSz9PAAS", "success"=>true, "errors"=>[]},
        "httpHeaders"=>{"Location"=>"/services/data/v47.0/sobjects/Account/0011k00000wSz9PAAS"},
        "httpStatusCode"=>201,
        "referenceId"=>"refAccount"},
       {"body"=>{"id"=>"0031k00000kBcbOAAS", "success"=>true, "errors"=>[]},
        "httpHeaders"=>{"Location"=>"/services/data/v47.0/sobjects/Contact/0031k00000kBcbOAAS"},
        "httpStatusCode"=>201,
        "referenceId"=>"refContactId"}]}>,
```

I noticed that the keys for the response object are all in camelcase with a lowercase letter at the beginning, but Restforce is [expecting camel caps case](https://github.com/restforce/restforce/blob/main/lib/restforce/concerns/composite_api.rb#L27-L33). Changing all of these locally seems to make the request results parse successfully (still testing all of the logic to be sure).

robob27:
Changes the capitalization expected in the compositeResponse object to match the docs here:

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/responses_composite.htm
https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_subrequest_result.htm

I also tested the changes locally both with and without `all_or_none: true` and the `CompositeAPIError` appears to be raised with the last error as expected.

I am unsure if this is an error in the initial implementation, or if there is some API version out there that uses this as is. The composite_api_spec tests are written with API version 38 (which is much older than when composite support was added to restforce, and might just be copy/paste from other specs?). I couldn't find documentation on this type of composite request in [the docs for version 38](https://developer.salesforce.com/docs/atlas.en-us.204.0.api_rest.meta/api_rest/intro_what_is_rest_api.htm). 

I did find documentation on this request in the [API version 39 docs](https://developer.salesforce.com/docs/atlas.en-us.206.0.api_rest.meta/api_rest/responses_composite.htm), and the response object appears to have the same type of capitalization in the response object as it does currently (API version 53), except for `referenceId` vs `referenceID` which we don't appear to extract from the response anyways. Adding these details because I couldn't really figure out if this is a breaking change or not.

Closes #678

### DIFF
--- a/lib/restforce/concerns/composite_api.rb
+++ b/lib/restforce/concerns/composite_api.rb
@@ -24,12 +24,12 @@ module Restforce
         }
         response = api_post('composite', properties.to_json)
 
-        results = response.body['CompositeResponse']
-        has_errors = results.any? { |result| result['HttpStatusCode'].digits.last == 4 }
+        results = response.body['compositeResponse']
+        has_errors = results.any? { |result| result['httpStatusCode'].digits.last == 4 }
         if all_or_none && has_errors
-          last_error_index = results.rindex { |result| result['HttpStatusCode'] != 412 }
+          last_error_index = results.rindex { |result| result['httpStatusCode'] != 412 }
           last_error = results[last_error_index]
-          raise CompositeAPIError, last_error['Body'][0]['errorCode']
+          raise CompositeAPIError, last_error['body'][0]['errorCode']
         end
 
         results
--- a/spec/unit/concerns/composite_api_spec.rb
+++ b/spec/unit/concerns/composite_api_spec.rb
@@ -128,7 +128,7 @@ describe Restforce::Concerns::CompositeAPI do
   describe '#composite' do
     let(:method) { :composite }
     let(:all_or_none) { false }
-    let(:response) { double('Faraday::Response', body: { 'CompositeResponse' => [] }) }
+    let(:response) { double('Faraday::Response', body: { 'compositeResponse' => [] }) }
     it_behaves_like 'composite requests'
   end
 
@@ -136,7 +136,7 @@ describe Restforce::Concerns::CompositeAPI do
     let(:method) { :composite! }
     let(:all_or_none) { true }
     let(:response) do
-      double('Faraday::Response', body: { 'CompositeResponse' => [] })
+      double('Faraday::Response', body: { 'compositeResponse' => [] })
     end
     it_behaves_like 'composite requests'
   end


timrogers:
This is a tricky one given that, in the PR that introduce this code, #528, there were people who said the code was working for them in production. I've tagged a couple of them to see what they say.

porbas:
I have the same issue here. Moreover, the capitalization proposed in this PR is aligned with examples from Salesforce docs: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_composite_upsert_example.htm

timrogers:
Let's give this merge a go. I'm sold given the docs and a couple of people with the issue. Let's see what happens.

porbas:
Thank you for merge!
It seems that build fails due to breaking change in Faraday: https://github.com/lostisland/faraday/pull/1367

PR #701, opened 2022-01-26
Add support for Faraday 1.9

magni-:
I have 1.9.3 installed locally and things seem to work fine? I'm less sure about 2.x, specifically the `faraday_middleware` [deprecation](https://github.com/lostisland/faraday_middleware#%EF%B8%8F-deprecation-warning-%EF%B8%8F):

> As highlighted in Faraday's [UPGRADING](https://github.com/lostisland/faraday/blob/main/UPGRADING.md) guide, `faraday_middleware` is DEPRECATED, and will not be updated to support Faraday 2.0. If you rely on `faraday_middleware` in your project and would like to support Faraday 2.0:
>
>    The `json` middleware (request and response) are now both bundled with Faraday 🙌
>    The `instrumentation` middleware is bundled with Faraday
>    All other middleware, they'll be re-released as independent gems compatible with both Faraday v1 and v2

Are there any other middleware that will need to be explicitly required other than `faraday-multipart` ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,5 @@ test/version_tmp
 tmp
 .rspec
 .ruby-version
+.tool-versions
 *.swp
--- a/lib/restforce/file_part.rb
+++ b/lib/restforce/file_part.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-begin
-  require 'faraday/file_part'
+%w[faraday/multipart faraday/file_part faraday/upload_io].find do |faraday|
+  require faraday
 rescue LoadError
-  require 'faraday/upload_io'
+  false
 end
 
 module Restforce
--- a/restforce.gemspec
+++ b/restforce.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.6'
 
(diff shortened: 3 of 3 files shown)


magni-:
Looks like [this StackOverflow answer](https://stackoverflow.com/a/34042681/1612744) doesn't work from a `.gemspec` file...  I guess the fix is to drop the last commit and then have a separate version supporting Faraday 2 only? Alternatively we could remove both `faraday_middleware` and `faraday-multipart` from the `.gemspec` file and instead add version validation to where these gems are required in `lib/restforce.rb` and `lib/restforce/file_part.rb`.

magni-:
Second attempt at Faraday 2.x support:
Commit: https://github.com/magni-/restforce/commit/de85538b0e708e202ea43e064949ba83121e6381
Failed build: https://app.circleci.com/pipelines/github/restforce/restforce/332/workflows/3337d13f-3bb3-43ee-a713-db6852793af5

This build failed because `faraday-multipart` was missing, since it's no longer in the `.gemspec` (the expectation was that users would be told to add the relevant middleware gem to their Gemfile depending on their version of `faraday`).

That's deep enough of a rabbit hole for today, so I'm content with leaving this PR being just 1.9 support.

timrogers:
Closing in favour of the now-merged #718 (based on this PR - thanks ❤️ )

PR #712, opened 2022-05-02
Follow redirects on authentication

nhocki:
This pull request adds the Faraday middleware to follow redirects when calling `authenticate!` because Salesforce will redirect to a classic domain (`.my.salesforce.com) when trying to use a lightning one (`.lightning.force.com`).

This is only needed when the client calls `authenticate!` directly to get a fresh token. Every other interaction happens through the `Connection` concern and the middleware is already there.

### DIFF
--- a/lib/restforce/config.rb
+++ b/lib/restforce/config.rb
@@ -91,29 +91,31 @@ module Restforce
       end
     end
 
-    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '26.0' }
+    option :api_version, default: lambda { ENV.fetch('SALESFORCE_API_VERSION', '26.0') }
 
     # The username to use during login.
-    option :username, default: lambda { ENV['SALESFORCE_USERNAME'] }
+    option :username, default: lambda { ENV.fetch('SALESFORCE_USERNAME', nil) }
 
     # The password to use during login.
-    option :password, default: lambda { ENV['SALESFORCE_PASSWORD'] }
+    option :password, default: lambda { ENV.fetch('SALESFORCE_PASSWORD', nil) }
 
     # The security token to use during login.
-    option :security_token, default: lambda { ENV['SALESFORCE_SECURITY_TOKEN'] }
+    option :security_token, default: lambda {
+                                       ENV.fetch('SALESFORCE_SECURITY_TOKEN', nil)
+                                     }
 
     # The OAuth client id
-    option :client_id, default: lambda { ENV['SALESFORCE_CLIENT_ID'] }
+    option :client_id, default: lambda { ENV.fetch('SALESFORCE_CLIENT_ID', nil) }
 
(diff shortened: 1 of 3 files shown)


nhocki:
Added a second commit that fixes Rubocop violations to get the checks green.

timrogers:
Thanks for the PR! Do you think we should consider passing the Lightning URL to be valid? Or should we just say that we expect a Salesforce URL, and potentially validate that?

nhocki:
👋  @timrogers, I think the lightning URL should be considered valid since it's a valid Salesforce URL and handling the redirect would make it similar to the other requests made from the `Connection` concern.

lib/restforce/config.rb
```diff
@@ -91,29 +91,31 @@ def option(*args)
       end
     end
 
-    option :api_version, default: lambda { ENV['SALESFORCE_API_VERSION'] || '26.0' }
+    option :api_version, default: lambda { ENV.fetch('SALESFORCE_API_VERSION', '26.0') }
```
timrogers, line 94: Nice improvements ✨

timrogers (commented):
Thanks for the PR! ❤️ My only slight concern with this is that there could be other Salesforce APIs that return redirects, and this change could affect the behaviour of those APIs. 

What do you think? One option, I guess, would be to write a simple custom Faraday middleware that only follows redirects for the authentication request.

timrogers:
The simplest way to do the above, I think, would be to create a class which inherits from the current middleware and defines its own `#follow_redirect?`. 

The risk of that approach is that the underlying middleware implementation changes in some future version. We can probably catch that with a good integration test.

nhocki:
@timrogers 👋 !

> My only slight concern with this is that there could be other Salesforce APIs that return redirects, and this change could affect the behaviour of those APIs.

This change would only affect calling `authenticate!` directly to get a new token. From what I understood from the gem, every other interaction goes through the `Connection` concern which is already following redirects. So if any other Salesforce API returns a redirect, we're already following it.

The other option would be to allow users of the gem to provide a block and yield the `builder` there so they can manually add new/custom middleware. This might be a great improvements even if this gets merged and we start following redirects.

timrogers:
> @timrogers 👋 !
> 
> > My only slight concern with this is that there could be other Salesforce APIs that return redirects, and this change could affect the behaviour of those APIs.
> 
> This change would only affect calling `authenticate!` directly to get a new token. From what I understood from the gem, every other interaction goes through the `Connection` concern which is already following redirects. So if any other Salesforce API returns a redirect, we're already following it.
> 
> The other option would be to allow users of the gem to provide a block and yield the `builder` there so they can manually add new/custom middleware. This might be a great improvements even if this gets merged and we start following redirects.

Right you are! 🏁 

I've merged these changes #717 - I created a new PR so I could cherry-pick and get GitHub Actions running.

PR #721, opened 2022-05-30
Add support for Faraday v2.x, and drop support for Faraday versions before v1.1.0

timrogers:
**This pull request adds support for Faraday v2.x, and drops support for Faraday versions before v1.1.0.** I plan to release this change as a **new major version, v6.0.0, starting with a release candidate version** to gather feedback. 

In the new major version, I will also **drop support for Ruby 2.6**, which has reached [EOL](https://www.ruby-lang.org/en/downloads/).

Finally getting support for Faraday v2.x is obviously a big win. It's something that people have been asking for. 

That makes sense, because Restforce being stuck on v1.x forces you to make a choice: do you drop Restforce, or do you stop updating other packages as they evolve and demand you use Faraday v2.x?

Clearly, dropping support for Faraday v0.x and the v1.0.x minor version isn't so positive, but I do believe that it's the right thing to do. There are a lot of differences between the Faraday versions, and for an open-source project with one active maintainer, it's a tonne of work and cognitive load to manage all of that. It increases the complexity of the code, and makes bugs much more likely.

What will be the first supported version of Faraday, v1.1.0, was [released](https://github.com/lostisland/faraday/releases/tag/v1.1.0) in October 2020, and the v0.x series hasn't been seriously updated since [January 2020](https://github.com/lostisland/faraday/releases/tag/v0.17.3).


albandum:
Hi, big blocker for us there (using in parallel a library that needs faraday >= 2.0
Any idea when this PR could be merged ?

Thanks a lot !

timrogers:
> Hi, big blocker for us there (using in parallel a library that needs faraday >= 2.0 Any idea when this PR could be merged ?
> 
> Thanks a lot !

I'd love to be able to merge it, but I've struggled with getting it to work. I have a bunch of failing tests, and I don't know why. Any help would be appreciated!

albandum:
> > Hi, big blocker for us there (using in parallel a library that needs faraday >= 2.0 Any idea when this PR could be merged ?
> > Thanks a lot !
> 
> I'd love to be able to merge it, but I've struggled with getting it to work. I have a bunch of failing tests, and I don't know why. Any help would be appreciated!

Thanks, I've just taken a look a little question before : the changes you made in lib/restforce/middleware/caching.rb are linked to the upgrade ? It seems to add a bunch of methods that didn't exist before ?

timrogers:
That caching bit is strange! I don't remember why I did it - I started on this PR a long time ago now. 

It'd definitely be worth trying to remove that and see if it fixed the tests. 

I'm certain that I didn't write that code myself. I definitely copied it from somewhere - perhaps its taken from an external library? Maybe GitHub Search can help to find where it came fro.

albandum:
Looks like you copy/pasted a few things from the Caching part of FaradayMiddleware.
Maybe a good option would be to try https://github.com/sourcelevel/faraday-http-cache ? 
I've tried a bit, but I'm getting a few other tests failing from reasons I can't understand yet. Here's the `caching.rb` I'm using if you want to take a look :

```# frozen_string_literal: true

module Restforce
  class Middleware::Caching < Faraday::HttpCache
    def call(env)
      expire(cache_key(env)) unless use_cache?
      super
    end

    def expire(key)
      cache&.delete(key)
    end

    # We don't want to cache requests for different clients, so append the
    # oauth token to the cache key.
    def cache_key(env)
      super(env) + hashed_auth_header(env)
    end

    def use_cache?
      @options[:use_cache]
    end

    def hashed_auth_header(env)
      Digest::SHA1.hexdigest(
        env[:request_headers][Restforce::Middleware::Authorization::AUTH_HEADER]
      )
    end
  end
end
```

Thanks a lot for your work, I'd love to be able to use Restforce, it looks really great !

timrogers:
@thatguysimon @albandum @mcfoton I can see that you've all reacted to this PR in one way or another. Would you be willing to give this code a try to check that it doesn't have any negative effects? 

All of the tests are passing, but I know that that doesn't necessarily mean 100% coverage.

timrogers:
...and the same goes to everyone else as well ✨ If you'd be willing to try this locally or in production, that would be amazing, so we can move Restforce into the future.

timrogers:
I'm going to merge this into a `v6.0.0.rc.1` branch and get ready to release it so people can try it easily.

timrogers:
We now have a discussion [here](https://github.com/restforce/restforce/discussions/733) calling for people to help test the new version.

mcfoton:
Awesome, we'll take it for a spin! Thanks for all the effort 🙏

albandum:
This is great, thanks so much ! Will take a look as soon as I can !

PR #740, opened 2022-08-12
Don't specify latest api version in docs

iloveitaly:
Makes the readme a bit more automatically up to date by not specifying an exact api version.


timrogers (approved):
❤️

PR #741, opened 2022-08-12
Add refresh method to sobject

iloveitaly:
I've found this to be a really helpful method to have on sobjects.

### DIFF
--- a/lib/restforce/sobject.rb
+++ b/lib/restforce/sobject.rb
@@ -16,6 +16,12 @@ module Restforce
       @client.describe_layouts(sobject_type, layout_id)
     end
 
+    def refresh
+      ensure_id
+      new_record = @client.find(sobject_type, self.Id)
+      deep_update(new_record.attrs)
+    end
+
     # Public: Persist the attributes to Salesforce.
     #
     # Examples


lib/restforce/sobject.rb
```diff
@@ -16,6 +16,12 @@ def describe_layouts(layout_id = nil)
       @client.describe_layouts(sobject_type, layout_id)
     end
 
+    def refresh
+      ensure_id
+      new_record = @client.find(sobject_type, self.Id)
+      deep_update(new_record.attrs)
```
timrogers, line 22: Why do you use `deep_update` rather than just returning the result of `#find`?

lib/restforce/sobject.rb
```diff
@@ -16,6 +16,12 @@ def describe_layouts(layout_id = nil)
       @client.describe_layouts(sobject_type, layout_id)
     end
 
+    def refresh
+      ensure_id
+      new_record = @client.find(sobject_type, self.Id)
```
timrogers, line 21: Are we confident that `#find` works with all `sobject_type`s? Is there anything we could do to validate that?

timrogers (changes requested):
This change is a really nice idea ✨ I've left some questions. Would you be up for writing a test too?

timrogers (changes requested):
This change is a really nice idea ✨ I've left some questions. Would you be up for writing a test too?

lib/restforce/sobject.rb
iloveitaly: I haven't found any object which it doesn't—is there a specific case you have in mind where you think we'd run into trouble?

lib/restforce/sobject.rb
iloveitaly: It allows for:

```
order.refresh
```

Without having to reassign:

```
order = order.refresh
```

Seemed a bit cleaner IMHO, but returning the new object as well is probably better:

```suggestion
      deep_update(new_record.attrs)
      new_record
```

iloveitaly:
@timrogers Answered the questions! Happy to write a test as well.

lib/restforce/sobject.rb
timrogers: Excellent point! That makes sense - it gives you ActiveRecord-like behaviour.

lib/restforce/sobject.rb
timrogers: I don't have any specific ideas, to. E honest. I'm happy to run with this, and I'm sure people will tell us if it doesn't work across the board. 

If it doesn't, you should get a sensible exception anyway.

lib/restforce/sobject.rb
```diff
@@ -16,6 +16,13 @@ def describe_layouts(layout_id = nil)
       @client.describe_layouts(sobject_type, layout_id)
     end
 
+    def refresh
```
timrogers, line 19: What do you think of renaming this to `#reload` to match ActiveRecord?

timrogers (commented):
I've suggested an alternative method name. With a test, I'd love to merge this ❤️

lib/restforce/sobject.rb
iloveitaly: I'm partial to `refresh` since it is what the Stripe & NetSuite API bindings use for this operation, but I'm happy to change it if you feel strong about the name!

timrogers: I'd like to use the ActiveRecord naming. I think it'll be more intuitive for most Ruby developers.

mbianco-stripe: Got it! Will change.

What do you think about aliasing `refresh` to the `reload` method?

lib/restforce/sobject.rb
iloveitaly: Got it! Will change.

What do you think about aliasing `refresh` to the `reload` method?

timrogers: ❤️ I'd prefer just to have one name to keep the surface area small. We can use both words when we document it in the readme to make it discoverable.

timrogers:
@iloveitaly Just a reminder to take a look at this if you get a chance - I'd love to get it merged ✨

PR #749, opened 2022-10-07
Add sObject Collections create call

igorlvicente:


dscataglini:
The rdoc are incorrect. Create and Update take an attribute key that specify what kind of sObject you're creating or updating.
So it should be more like this:
```
client.create_collection(
  [{"attributes" : {"type" : "Account"}, Name: 'Foo Inc.'}, 
   {"attributes" : {"type" : "Account"}, Name: 'Bar Corp.'}])
```
Same thing for update
```
client.update_collection!([
   {"attributes" : {"type" : "Account"}, Id: '1', Name: 'Foo Inc.'}, 
   {"attributes" : {"type" : "Account"}, Id: '2', Name: 'Bar Corp.'}
])
```

from the documentation https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_create.htm
```
POST /composite/sobjects
{
   "allOrNone" : false,
   "records" : [{
      "attributes" : {"type" : "Account"},
      "Name" : "example.com",
      "BillingCity" : "San Francisco"
   }, {
      "attributes" : {"type" : "Contact"},
      "LastName" : "Johnson",
      "FirstName" : "Erica"
   }]
}
```

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_update.htm
```
PATCH /composite/sobjects
{
   "allOrNone" : false,
   "records" : [{
      "attributes" : {"type" : "Account"},
      "id" : "001xx000003DGb2AAG",
      "NumberOfEmployees" : 27000
   },{
      "attributes" : {"type" : "Contact"},
      "id" : "003xx000004TmiQAAS",
      "Title" : "Lead Engineer"
   }]
}
```

This is because you can create/update different kind of objects in the same request

PR #758, opened 2022-11-10
Allow Faraday v2.x

marcrohloff:
Faraday is currently at v2.6 but is currently locked at <= 2.4


timrogers:
Thanks for spotting this ❤️ I'm going to close this in favour of #758 because there is a reason why we choose to enable support for new Faraday minor releases one at a time. But thank you for taking the time to contribute!

PR #765, opened 2022-12-08
include response body for CompositeAPIError

shravan097:
Related to https://github.com/restforce/restforce/issues/766
### Changes
Allow `CompositeAPIError` to include Faraday's response
### Testing
I added one test related to this change &  ran rspec locally.

### DIFF
--- a/lib/restforce.rb
+++ b/lib/restforce.rb
@@ -50,15 +50,6 @@ module Restforce
   APIVersionError     = Class.new(Error)
   BatchAPIError       = Class.new(Error)
 
-  class CompositeAPIError < Error
-    attr_reader :composite_response
-
-    def initialize(msg, composite_response = nil)
-      @composite_response = composite_response
-      super(msg)
-    end
-  end
-
   # Inherit from Faraday::ResourceNotFound for backwards-compatibility
   # Consumers of this library that rescue and handle Faraday::ResourceNotFound
   # can continue to do so.
@@ -68,6 +59,7 @@ module Restforce
   # Consumers of this library that rescue and handle Faraday::ClientError
   # can continue to do so.
   ResponseError       = Class.new(Faraday::ClientError)
+  CompositeAPIError   = Class.new(ResponseError)
   MatchesMultipleError= Class.new(ResponseError)
   EntityTooLargeError = Class.new(ResponseError)
 
--- a/lib/restforce.rb
+++ b/lib/restforce.rb
@@ -49,7 +49,15 @@ module Restforce
   UnauthorizedError   = Class.new(Faraday::ClientError)
   APIVersionError     = Class.new(Error)
   BatchAPIError       = Class.new(Error)
-  CompositeAPIError   = Class.new(Error)
+
+  class CompositeAPIError < Error
+    attr_reader :composite_response
+
+    def initialize(msg, composite_response = nil)
+      @composite_response = composite_response
+      super(msg)
+    end
+  end
 
   # Inherit from Faraday::ResourceNotFound for backwards-compatibility
   # Consumers of this library that rescue and handle Faraday::ResourceNotFound
--- a/lib/restforce/concerns/composite_api.rb
+++ b/lib/restforce/concerns/composite_api.rb
@@ -29,7 +29,7 @@ module Restforce
         if all_or_none && has_errors
           last_error_index = results.rindex { |result| result['httpStatusCode'] != 412 }
           last_error = results[last_error_index]
-          raise CompositeAPIError, last_error['body'][0]['errorCode']
+          raise CompositeAPIError.new(last_error['body'][0]['errorCode'], results)
         end
 
         results
--- a/lib/restforce/concerns/composite_api.rb
+++ b/lib/restforce/concerns/composite_api.rb
@@ -29,7 +29,7 @@ module Restforce
         if all_or_none && has_errors
           last_error_index = results.rindex { |result| result['httpStatusCode'] != 412 }
           last_error = results[last_error_index]
-          raise CompositeAPIError.new(last_error['body'][0]['errorCode'], results)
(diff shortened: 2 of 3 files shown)


lib/restforce/concerns/composite_api.rb
```diff
@@ -23,22 +23,21 @@ def composite(all_or_none: false, collate_subrequests: false)
           collateSubrequests: collate_subrequests
         }
         response = api_post('composite', properties.to_json)
+        response.body['compositeResponse']
```
jeffersonhwang, line 26: Isn't the purpose of this PR to return the entire response body when raising an error? This change doesn't seem to solve that issue as line 36 hasn't changed.

shravan097: there were 3 ways to go about this:
1. Return `CompositeAPIError` with entire error message as a string. The error message is an Array of Hash here so it needs to be serialized into string before forming `CompositeAPIError`
   - **Tradeoff**:  Error message needs to be serialized  by SDK & de-serialized by Client.
2. Extend `CompositeAPIError` to intake extra field to store error message.
   - **Tradeoff** : `CompositeAPIError` is now different error from rest of the other error object which is inconsistent. 
3. Just return entire `compositeResponse` in `complaint` & throw `CompositeAPIError` in `complaint!` if `allOrNone` is true & `compositeResponse` has error.
   -  **Tradeoff**:  Client needs to process error on its own when using `complaint` or rely on `complaint!` to throw error. 
 
With all the approach listed above, I personally feel approach 3 is cleaner. Every use case will be different and having the option to return a raw `compositeResponse` opens up client to their interpretation of when to raise the error.

shravan097: sorry I just realized a major flaw with approach 3 - it is a major breaking change to existing SDK users. I think approach 2 is the best way forward without breaking the existing functionality. I will update this PR.

lib/restforce.rb
```diff
@@ -49,7 +49,15 @@ module Tooling
   UnauthorizedError   = Class.new(Faraday::ClientError)
   APIVersionError     = Class.new(Error)
   BatchAPIError       = Class.new(Error)
-  CompositeAPIError   = Class.new(Error)
+
+  class CompositeAPIError < Error
```
jeffersonhwang, line 53: We should figure out the latest pattern in the Restforce library. I see some custom errors extend [Faraday Errors](https://github.com/lostisland/faraday/blob/main/lib/faraday/error.rb) which gives you an underlying response attribute. This line in [raise_error.rb](https://github.com/restforce/restforce/blob/10b22e6b10b7cca876d84b1ea2734d0478b4720b/lib/restforce/middleware/raise_error.rb#LL33C50-L33C50) also seems to suggest response_values may already be attached to errors

shravan097: good find. I can simply do `CompositeAPIError   = Class.new(Faraday::ClientError)` and attach response as a second param. As for the [raise_error.rb](https://github.com/restforce/restforce/blob/10b22e6b10b7cca876d84b1ea2734d0478b4720b/lib/restforce/middleware/raise_error.rb#LL33C50-L33C50) & Restforce middleware, Composite Response returns 200 even when there is a 400 response for subrequest so it skips this middleware and is directly handled on [composite_api.rb](https://github.com/restforce/restforce/pull/765/files#diff-652b9108b81ff7406ad380bfc726d290ecc4bd37613f0a83dfc18d0b7d3abd33R32).

lib/restforce.rb
```diff
@@ -49,7 +49,7 @@ module Tooling
   UnauthorizedError   = Class.new(Faraday::ClientError)
   APIVersionError     = Class.new(Error)
   BatchAPIError       = Class.new(Error)
-  CompositeAPIError   = Class.new(Error)
+  CompositeAPIError   = Class.new(Faraday::ClientError)
```
jeffersonhwang, line 52: I want to say `ResponseError` may be more appropriate. See lines 62 to 64.

spec/unit/concerns/composite_api_spec.rb
```diff
@@ -139,5 +139,31 @@
       double('Faraday::Response', body: { 'compositeResponse' => [] })
     end
     it_behaves_like 'composite requests'
+
+    it 'has response in CompositeAPIError' do
```
jeffersonhwang, line 143: Should this go with the shared examples? This case can happen whenever `composite` is called with `allOrNone: true` - not exclusive to `composite!`.

lib/restforce/concerns/composite_api.rb
```diff
@@ -29,7 +29,7 @@ def composite(all_or_none: false, collate_subrequests: false)
         if all_or_none && has_errors
           last_error_index = results.rindex { |result| result['httpStatusCode'] != 412 }
           last_error = results[last_error_index]
-          raise CompositeAPIError, last_error['body'][0]['errorCode']
+          raise CompositeAPIError.new(last_error['body'][0]['errorCode'], response)
```
jeffersonhwang, line 32: Should this be `response` or `results`?

jeffersonhwang:
@shravan097 Would also recommend tying this to an existing issue or creating a new one

lib/restforce/concerns/composite_api.rb
shravan097: this should be Faraday's `response` because it is inheriting Faraday::ClientError. [reference Faraday's code](https://github.com/lostisland/faraday/blob/main/lib/faraday/error.rb#L32)

spec/unit/concerns/composite_api_spec.rb
shravan097: you are right. forgot to update this.

shravan097:
@timrogers hi Tim! Let me know if this does not make sense. thanx

timrogers (approved):

timrogers:
@shravan097 Sorry for the delay in taking a look at this! It looks great.

PR #779, opened 2023-01-18
When a response claims to be gzipped but isn't, return the body as it is and don't explode

### LINKED ISSUE #761: Error "not in gzip format" calling `Restforce.find`
From v6.1.0 I am getting this error when calling `.find`. Other methods like `.create` are working ok.

```rb
@client = Restforce.new(
  username: "",
  client_id: "",
  instance_url: "",
  jwt_key: Base64.urlsafe_decode64(""),
  api_version: "41.0"
)

@client.find("Lead", "lead_id")
# => /app/vendor/bundle/ruby/3.1.0/gems/restforce-6.1.0/lib/restforce/middleware/gzip.rb:30:in `initialize': not in gzip format (Zlib::GzipFile::Error)
```

Restforce version: 6.1.0

timrogers:
In the past couple of months (see #761), users have reported an issue where some API calls raise a `Zlib::GzipFile::Error`.

The root cause [seems][1] to be that Restforce is seeing a `Content-Encoding` response header and thinking that it needs to decompress the body, when in fact, the body is not compressed. When you try to decompress something with Gzip that isn't compressed, it raises an error.

This catches that error and returns the body as is.

It doesn't really answer the underlying question - "why do we think the response is encoded?" - but it should unblock users.

Fixes #761.

[1]: https://github.com/restforce/restforce/issues/761#issuecomment-1329012277

### DIFF
--- a/lib/restforce/middleware/gzip.rb
+++ b/lib/restforce/middleware/gzip.rb
@@ -28,6 +28,10 @@ module Restforce
     # Internal: Decompresses a gzipped string.
     def decompress(body)
       Zlib::GzipReader.new(StringIO.new(body)).read
+    rescue Zlib::GzipFile::Error
+      # We thought the response was gzipped, but it wasn't. Return the original
+      # body back to the caller. See https://github.com/restforce/restforce/issues/761.
+      body
     end
   end
 end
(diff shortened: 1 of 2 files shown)


timrogers:
@schm has [validated](https://github.com/restforce/restforce/issues/761#issuecomment-1386858306) that this works!

PR #833, opened 2023-10-05
Register JSON middleware as restforce_json to avoid overwriting default

dbackeus:
Since the gem was registering it's own custom JSON middleware as `:json` it could end up taking precedence over the default Faraday JSON middleware in application code and break things.

This PR changes the registration name to `:restforce_json` to avoid any chance of collision.

### DIFF
--- a/lib/restforce/concerns/connection.rb
+++ b/lib/restforce/concerns/connection.rb
@@ -33,7 +33,7 @@ module Restforce
           # Handles multipart file uploads for blobs.
           builder.use Restforce::Middleware::Multipart
           # Converts the request into JSON.
-          builder.request :json
+          builder.request :restforce_json
           # Handles reauthentication for 403 responses.
           if authentication_middleware
             builder.use authentication_middleware, self, options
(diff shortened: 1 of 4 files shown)


timrogers (approved):

timrogers:
Change makes total sense - thanks ❤️ I can imagine this one would have been hard to debug...

Did you have your own custom middleware with this name? I'm interested to know where the issue emerged.

dbackeus:
Yes, it was quite the debugging conundrum 😅 

In our app the restforce middleware took precedence over the one from the `faraday_middleware` gem which had a subtle difference in that it will parse response body as JSON even if no explicit `Content-Type: application/json` is returned in the response. This broke a whole bunch of WebMocked specs which didn't include mocking of the content-type header.

Overall I come out of this feeling like the `register_middleware` idea based on arbitrary symbols and constant autoloading on the Faraday side was a mistake. Settling for `builder.response Full::Path:To:MiddlewareConstant` would have been super simple to implement and comprehend as well as had symmetry with other middleware API's like Rack.

Thanks for approving 🙏

Issue #869, opened 2024-03-19
There's no way to know if `upsert` performed a `creation` or an `update`


FelixNumworks:
In the README, it's said:

> The upsert method will return the record Id if included in the response body from the Salesforce API; otherwise, it returns true. Currently the Salesforce API only returns the Id for newly created records.

As mentionned in [this Salesforce doc](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm), `upsert` now always returns the "id" along with a "created" attribute.

Thus, Restforce's upsert now always returns the id of the record in Salesforce, which makes it impossible to know if it was created or updated.
Also this makes the README deprecated.

Related to https://github.com/restforce/restforce/issues/556

Issue #556, opened 2020-07-16
"created" in upsert response isn't accessible


jiajiawang:
In API version 46.0 and later, upsert would always return Id on bother update and insert, along with {"created": true/false} which can tell it's an update or insert.

At the moment, upsert method return id only. Can "created" be included in the return?

dinshaw-he:
From the docs, the current successful response is 
```
{
    "id" : "00190000001pPvHAAU",
    "errors" : [ ],
    "success" : true,
    "created": true
}
```
It would be great to have the whole response returned. Currently one needs to use `upsert!` to surface any errors. If there is support for this I may be able to pitch in.

Issue #253, opened 2016-08-04
upsert on update not returning Id


theog:
An upsert using this library will only return bool value on an update. When using an external ID as a reference to the SalesForce record, it would be very useful to have the Id returned regardless of create or update would save an additional call to obtain the record. SalesForce REST API docs returns the Id. 

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm

timrogers:
This makes sense to me. Do you want to make a PR? I might not merge it in for a while as it'll be a breaking change.

timrogers:
@theog Ping! 🔔

anacaet:
Hy @theog and @timrogers any update on this? Would be great with this library has this option

timrogers:
Hi Ana,

Do you want to try making a pull request? I'd be happy with this change.

Tim

pangolingo:
I was looking for this functionality too. But according to Salesforce documenation (https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm) when a record is updated, the API only returns the updated fields. It does not return the Salesforce ID.

I don't see how we can fix this without doing an extra API call.

rgraff:
Confirming what pangolingo said. If a new record is created, the API returns a 200 with the newly created id in the body.  If a record is updated, the API simply returns a 204 with an empty body.

Restforce (verified in 4.2.1) does the best it can given the API limitation. Upsert will return the new id when a new record is created. When a new record is not created, it simply returns true. I'll do a PR to add this to the readme which might save people time in the future.

9re:
@pangolingo @rgraff 

As time has passed, looks like there is an update on Salesforce API side. I'm very happy if this project also follow this spec.

https://github.com/restforce/restforce/issues/556

FelixNumworks:
This fixed itself as the id is now returned by Salesforce. So "upsert" now always returns the id, for a creation or an update

PR #870, opened 2024-03-19
Modify upsert to return the "created" parameter along the "id"

FelixNumworks:
Fix https://github.com/restforce/restforce/issues/556
Fix https://github.com/restforce/restforce/issues/869

Related to https://github.com/restforce/restforce/issues/253

I didn't update the `spec` files as they seem already out-of-date

I'm not sure this is how you want to handle this.

### DIFF
--- a/lib/restforce/concerns/api.rb
+++ b/lib/restforce/concerns/api.rb
@@ -338,8 +338,9 @@ module Restforce
       #   # Update the record with external ID of 12
       #   client.upsert('Account', 'External__c', External__c: 12, Name: 'Foobar')
       #
-      # Returns true if the record was found and updated.
-      # Returns the Id of the newly created record if the record was created.
+      # Returns a hash with the 'id' of the Salesforce record, and a 'created'
+      # attribute, equal to true if the record was created, false if it was updated.
+      # ex: { 'id' => '0016000000MRatd', 'created' => false }
       # Returns false if something bad happens (for example if the external ID matches
       # multiple resources).
       def upsert(*args)
@@ -359,8 +360,9 @@ module Restforce
       #   # Update the record with external ID of 12
       #   client.upsert!('Account', 'External__c', External__c: 12, Name: 'Foobar')
       #
-      # Returns true if the record was found and updated.
-      # Returns the Id of the newly created record if the record was created.
+      # Returns a hash with the 'id' of the Salesforce record, and a 'created'
+      # attribute, equal to true if the record was created, false if it was updated.
+      # ex: { 'id' => '0016000000MRatd', 'created' => false }
       # Raises an exception if an error is returned from Salesforce, including the 300
       # error returned if the external ID provided matches multiple records (in which
       # case the conflicting IDs can be found by looking at the response on the error)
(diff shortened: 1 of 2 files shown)


FelixNumworks:
Great ! Thanks for you fast response.

Do you need me to change something somewhere to make it clear it's a breaking change (like in the commit name ?)

About tests, here the at the last line there is an `expect(result).to be true`

https://github.com/restforce/restforce/blob/5bfeb898ae9a4e82d1847a1aed77ee9d12946b90/spec/unit/concerns/api_spec.rb#L351-L360

I don't understand why it works, since the result is in fact a string:
(Maybe it just checks if `expect(result)` is not nil ?)

There a few others similar but this exemplify all of them.

timrogers:
@FelixNumworks Thanks! ❤️ I will take a look at this and the tests over the next few days. I might suggest a slightly different design for the `#upsert` method, but I'd like to have something you can test by the end of the week.

lib/restforce/concerns/api.rb
```diff
@@ -383,7 +385,7 @@ def upsert!(sobject, field, attrs)
                       "#{ERB::Util.url_encode(external_id)}", attrs
           end
 
-        response.body.respond_to?(:fetch) ? response.body.fetch('id', true) : true
+        response.body.slice("id", "created")
```
timrogers, line 388: There are some Salesforce API versions where this API returns `204 No Content` after an update ([source](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_upsert.htm)):

> In API version 45.0 and earlier, the HTTP status code is 204 (No Content) and there isn’t a response body.

I think using `#slice` will break for those API versions 😦

timrogers (changes requested):
This change is pretty tricky because there are lots of different Salesforce API versions with different behaviour that we probably want to continue to support.

I'm going to need to spend some more time on it and test across the versions if we want to make a change here.

lib/restforce/concerns/api.rb
FelixNumworks: I see :/ To be honest, I'm not very familiar with Salesforce, even less with old versions of it.

Maybe a solution could be:

```rb
# Update response for old API implementations that aren't returning the id
if !response.body.respond_to?(:fetch) || !response.body.has_key?("id")
  return { "created" => false }
end
# Create response for old API implementations that aren't returning "created"
if !response.body.has_key?("created")
  return { "id" => response.body["id"], "created" => true }
end
# New API
response.body.slice("id", "created")
```

But it becomes a bit complicated ^^"

Another possibility is to stay with the same Restforce spec (return true in case of Update, id in case of Create), like this:

```rb
# Update case for old API returning nothing
return true if !response.body.respond_to?(:fetch)
# Update case for old API returning "true"
return true if !response.body.has_key?("id")
# Update case for new API returning "created"
return true if response.body.has_key?("created") && response.body["created"] == false
# Create case
response.body["id"]
```

This last solution would solve https://github.com/restforce/restforce/issues/869 and https://github.com/restforce/restforce/issues/556 but not https://github.com/restforce/restforce/issues/253

PR #947, opened 2025-10-16
correct version constraint for faraday-follow_redirects to allow upgrade

espen:


fatkodima:
Can be closed - was fixed in #951.

